PR #1, opened 2017-07-05
Header component (with mask)

cathrholm:
Header component done; with the use of rem function and css media queries with em function.

### DIFF
--- a/src/components/molecules/Header/Header.html.TODO
+++ b/src/components/molecules/Header/Header.html.TODO
@@ -1,0 +1,9 @@
+<header class="header header--with-mask header--with-content-overlap">
+    <img class="header__icon" src="/public/icons/ico_favourites.svg" role="presentation" />
+    <span class="header__subtitle ">Pre-title</span>
+    <h1 class="heading heading--level-1">Page Title</h1>
+    <section class="header__preamble rich-text-area">
+        Preamble; an introductory and expressionary statement in a document
+        that explains the document's purpose and underlying philosophy
+    </section>
+</header>
--- a/src/components/molecules/Header/Header.jsx
+++ b/src/components/molecules/Header/Header.jsx
@@ -1,0 +1,24 @@
+import React from 'react';
+import Heading from '../../atoms/Heading/Heading';
+
+function classNames(withMask, withContentOverlap) {
+    const classNames = ['header'];
+    if (withMask) {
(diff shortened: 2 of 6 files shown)


src/components/molecules/Header/Header.jsx
```diff
@@ -0,0 +1,23 @@
...
+        classNames.push(`header--with-mask`)
+    }
+    if (withContentOverlap) {
+        classNames.push(`header--with-content-overlap`)
+    }
+
+    return classNames.join(' ');
+}
+
+const Header = ({iconUrl, runningTitle, pageTitle, preamble, withMask, withContentOverlap}) =>
```
alexyork: Can you add spaces between the JavaScript object?

Good: `{ foo }`
Bad: `{foo}`

src/components/molecules/Header/Header.jsx
```diff
@@ -0,0 +1,23 @@
...
+    <header className={classNames(withMask, withContentOverlap)}>
+        <img className="header__icon" src={iconUrl} role="presentation" />
+        <span className="header__subtitle ">{runningTitle}</span>
+        <h1 className="heading heading--level-1">{pageTitle}</h1>
```
alexyork: Can we use the `<Heading>` component here, to render the `<h1>`? Molecules should reuse smaller atoms where they can, to facilitate reuse :-)

src/components/molecules/Header/Header.html.TODO
```diff
@@ -0,0 +1,9 @@
+<header class="header header--with-mask header--with-content-overlap">
+    <img class="header__icon" src="/public/icons/ico_favourites.svg" role="presentation" />
+    <span class="header__subtitle ">Pre-title</span>
+    <h1 class="heading heading--level-1">Page Title</h1>
+    <section class="header__preamble rich-text-area">
```
alexyork: Remove `rich-text-area`. EpiServer may want to include that class in various places, but it isn't a styleguide concern (yet).

src/components/molecules/Header/Header.jsx
```diff
@@ -0,0 +1,23 @@
...
+        <section className="header__preamble rich-text-area">{preamble}</section>
```
alexyork: Remove `rich-text-area`. Same reason as above.

PR #2, opened 2017-07-06
Feature/pageheader menu

alexyork:
I wouldn't say this is 100% finished, or perfect, but it is a good enough V1, so I think we can get it into the codebase and iterate on it as and when needed.


src/components/molecules/PageHeader/PageHeader.jsx
```diff
@@ -10,41 +10,82 @@ import React from 'react';
...
+                    </div>
+                    <div className="page-header__search">
+                        <input className="page-header__search-input" type="text" placeholder="Spøk, neida... søk" />
+                    </div>
+                    <nav className="page-header__menu-nav">
+                        <ul className="page-header__menu-item-list">
+                            {_.map(this.props.menuLinks, (menuLink) =>
+                                <li className="page-header__menu-item">
+                                    <a className="page-header__menu-item-link" href={menuLink.url}>
+                                        {menuLink.icon ? <img className="page-header__menu-item-icon" src={menuLink.icon} alt="" role="presentation" /> : null}
```
cathrholm: There are several empty alt-attributes, should these be filled in later? 
And when an image is marked with `role="presentation"`, you don't need the alt-attribute because these elements are ignored by assistive technologies.

src/components/molecules/PageHeader/PageHeader.jsx
```diff
@@ -10,41 +10,82 @@ import React from 'react';
...
+                <div className="page-header__site-nav">
+                    <button className="page-header__menu-button page-header__icon-box" onClick={this.openMenu} aria-expanded={this.state.menuIsExpanded} aria-controls={this.props.menuId || "page-header-menu"} aria-pressed={this.state.menuIsExpanded}>
+                        <img className="page-header__menu-button-image page-header__icon-box-icon" src="/public/icons/ico_menu_purple.svg" alt="" role="presentation" />
+                        <span className="page-header__icon-box-text">Meny</span>
+                    </button>
+                </div>
+                <div className="page-header__site-tools">
+                    {cartShouldBeShown ?
+                        <a className="page-header__cart-button" href="#">
+                            <img className="page-header__cart-button-image" src="/public/icons/ico_buy_purple.svg" alt="Handlekurv" />
```
cathrholm: We should be consistent with the icons, but I'm not sure what the best solution is.
Here the cart have alt-text, so the screenreaders will pick up on this element. But on line 81, the user-icon has rol presentation and will not be picked up by screen readers.

src/components/molecules/PageHeader/PageHeader.jsx
```diff
@@ -10,41 +10,82 @@ import React from 'react';
  * It contains the Telia logo, the main navigation, and tools such as log-in info
  * and shopping cart.
  */
-const PageHeader = ({ logoUrl, logoTitle, logoImageDesktopPath, logoImageMobilePath, cartItemCount, isLoggedIn, loggedInUserName }) => {
-    const cartShouldBeShown = (cartItemCount > 0);
-    return (
-        <header className={ClassNames('page-header', { 'page-header--with-cart': cartShouldBeShown })}>
```
alexyork: Yes it is a function, but it is also a library we import:

```
import ClassNames from 'classnames';
import React from 'react';
```

There is no established naming conventions for imported libraries. Some have uppercased names, some don't. Here I used uppercase, but you're right, it could have been lower-cased :-)

src/components/molecules/PageHeader/PageHeader.jsx
```diff
@@ -10,41 +10,82 @@ import React from 'react';
...
+            <header className={ClassNames('page-header', { 'page-header--with-cart': cartShouldBeShown, 'page-header--menu-expanded': this.state.menuIsExpanded })}>
+                <div className="page-header__menu" id={this.props.menuId || "page-header-menu"}>
+                    <div className="page-header__close-menu">
+                        <button className="page-header__close-menu-button page-header__icon-box" onClick={this.closeMenu} aria-expanded={this.state.menuIsExpanded} aria-controls={this.props.menuId || "page-header-menu"} aria-pressed={!this.state.menuIsExpanded}>
+                            <img className="page-header__icon-box-icon" src="/public/icons/ico_delete_purple.svg" alt="" role="presentation" />
+                            <span className="page-header__icon-box-text">Lukk</span>
+                        </button>
+                    </div>
+                    <div className="page-header__search">
+                        <input className="page-header__search-input" type="text" placeholder="Spøk, neida... søk" />
```
alexyork: Good. Please fix for me!

src/components/molecules/PageHeader/PageHeader.jsx
alexyork: Good. Please fix for me!

src/components/molecules/PageHeader/PageHeader.jsx
alexyork: Good. Please fix for me!

PR #12, opened 2017-09-19
donut and progress charts

mkmarek:


src/components/atoms/ProgressBar/ProgressBar.jsx
```diff
@@ -0,0 +1,46 @@
...
+    max: PropTypes.number.isRequired,
+    barColor: PropTypes.string,
+    disabled: PropTypes.bool
+}
+
+ProgressBar.defaultProps = {
+    barColor: 'green'
+}
+
+export default ProgressBar
```
alexyork: All JS lines should end with semi-colons. I know this is optional in JS, but we have simply too much code by now to go for the no semi-colons, and now the best thing to do is be consistent. Thanks!

src/components/atoms/Spinner/Spinner.jsx
```diff
@@ -0,0 +1,43 @@
...
+Spinner.propTypes = {
+    className: PropTypes.string,
+    type: PropTypes.string
+}
+
+Spinner.defaultProps = {
+    type: 'lg'
+}
+
+export default Spinner
```
alexyork: Semi-colons.

src/components/molecules/DonutChart/DonutChart.jsx
```diff
@@ -0,0 +1,108 @@
...
+    loadingCaption: "Henter data...",
+    valueCaption: "gjenstår",
+    thresholds: [
+        { to: 0.2, type: 'red' },
+        { to: 0.6, type: 'orange' },
+        { to: 1, type: 'green' }
+    ]
+}
+
+export default DonutChart
```
alexyork: Semi-colon.

src/components/molecules/ProgressChart/ProgressChart.jsx
```diff
@@ -0,0 +1,82 @@
...
+    loadingCaption: "Henter data...",
+    valueCaption: "gjenstår",
+    thresholds: [
+        { to: 0.2, type: 'red' },
+        { to: 0.6, type: 'orange' },
+        { to: 1, type: 'green' }
+    ]
+}
+
+export default ProgressChart
```
alexyork: Semi-colons.

src/components/molecules/ProgressChart/ProgressChart.pcss
```diff
@@ -0,0 +1,59 @@
...
+        max-height: 20px;
+
+        top: 0;
+        padding: 0 1em;
+        display: flex;
+        position: relative;
+
+        text-align: center;
+        line-height: 1.3rem;
+        white-space: nowrap;
```
alexyork: Is there a reason behind this CSS grouping...?

src/components/colors.pcss
```diff
@@ -39,4 +41,5 @@
     --red: #FF3264;
     --dark-red: #FF2364;
     --light-red: #FF6464;
+    --venetian-red: #D0021B;
```
alexyork: I will double-check what we do when we are presented with new colours. The ones we already have come from the Telia Company brand, and we should try and use them as much as possible. I will get back to you :-)

src/components/molecules/ProgressChart/ProgressChart.pcss
mkmarek: After I had a second look it was pretty nonsensical. However I tried to separate styles that define different aspects of a block. Now I separated it into

- stuff that influences the element display properties (position, display)
- stuff that defines area around the content (padding, margin, top, bottom, ...)
- stuff that defines text (line-height, font-size, text-align, ...)

Not sure if you're using different approaches, or none at all but I can adjust.

PR #13, opened 2017-09-19
Feature/about telia

cathrholm:
Added these components; body-text-container (we had this already in epi-project), Numberboxes, StepByStep. Several sample pages of Information Article Page is also added.


alexyork:
No comments. This is pretty much perfect. Merging!

PR #16, opened 2017-09-28
Feature/pebbels on page

cathrholm:


alexyork:
Hi. Great work. As we discussed, lets use fixed numbers of pixels from the top of the page, not percentages, so that very long pages also look good.

jimmythesaint82:
Demo Page added
Pebbles have hardcoded position
Clickthrough bug removed

cathrholm:
Increased z-index to 1, to make the pebbles lay on top of table-element.  Also setting pointer-events back to auto for every element inside the class .page-pebbles. Good to go, merging to master!

PR #28, opened 2017-10-23
Feature/video

cathrholm:
- Video component.
- Refactored caption into it's own component that is used in Image, FullWidthImage and Video.
- Added a new example for Header component with video.
- Added a new Blog sample page.


src/components/atoms/Video/Video.html
```diff
@@ -0,0 +1,20 @@
+<!--
+    URL parameters with value 0/1 (true/false), showing default value:
+    * Suggest related videos after current video is finished: rel=1
+    * Show video player controls: controls=1
+    * Display information like the video title and uploader before the video starts playing: showinfo=1
+    * Autoplay: autoplay=0
+-->
```
cathrholm: Should remove these comments and add the relevant information in the documentation instead.
Just wanted to write this down to get some clarification on how the videos should behave.

PR #29, opened 2017-10-31
Speech bubbles on subscriptions

alexyork:


src/components/molecules/Subscription/Subscription.jsx
```diff
@@ -63,6 +63,9 @@ export default class Subscription extends React.Component {
                 </section>
                 {(this.state.isExpanded || this.props.isShowingFeatures) ?
                     <section className="subscription__features">
+                        {this.props.speechBubbleText
+                            ? <div className="subscription__speech-bubble">{this.props.speechBubbleText}</div>
+                            : <div className="subscription__speech-bubble subscription__speech-bubble--empty"></div>}
```
cathrholm: I can see why this is done, but in general I don't think we should render html if it's not supposed to be visual. Screen readers read elements that has visibility hidden.

alexyork: I agree, it's a small hack. How can we solve this in another way? We want the feature to be in the same place regardless of what is above it. I tried absolute positioning but it was tricky. Do you wanna fix? :-)

cathrholm: Maybe we can fix this when we refactor subscription to use the Box component? I think we'll have to make adjustments anyway at that point :-)

alexyork: Ja!

PR #30, opened 2017-10-31
Feature/product promo

cathrholm:
I would like to go through this component with the designer, to get feedback on the correct font-sizes, spacing between content in the box and breakpoint between mobile and desktop, but for now I think it's good enough to get this out.


alexyork:
Standalone boxes with `--small` and `--medium` have some side margin. As we discussed, standalone boxes should have no margin on the sides. Maybe they should have margin-bottom though!

src/components/atoms/Box/Box.jsx
```diff
@@ -39,18 +39,23 @@ export default class Box extends React.Component {
             this.closeBoxClick = this.closeBoxClick.bind(this);
         }
     }
-    boxContainerClick() {
-        if (!this.state.isExpanded) {
-            this.setState({ isExpanded: true });
+    boxContainerClick(e) {
```
cathrholm: @alexyork, can you have a quick look at this change?
I added tabIndex and keyUp eventlistener to the Box component as well.

PR #33, opened 2017-11-27
Feature/content feedback

cathrholm:
- content feedback component
- added the component on all pages in the styleguide and posting feedback to slack channel '#styleguide'
- refactored slack.js to make use of functionality to post to slack from several places (when deploying and on content feedback in styleguide)


alexyork:
1. Remove `content-feedback__button` class and associated CSS.
2. Use `button button--small`
3. Edit the button component: margin right, no margin bottom.
4. Add a modifier to button (use the same naming conventions that `container` uses) for a special case where you want margin-bottom.

PR #43, opened 2017-12-12
Label component first version

cathrholm:
We need other form components to be done as well before we say that this component is finished, because we need to show some examples of Label with a form element.

### DIFF
--- a/src/components/atoms/Label/Label.jsx
+++ b/src/components/atoms/Label/Label.jsx
@@ -1,0 +1,38 @@
+import React from 'react';
+
+function classNames(isTextBox, formTypeClassName) {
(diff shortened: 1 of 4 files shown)


src/components/atoms/Label/Label.jsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+
+function classNames(isTextBox, formTypeClassName) {
+    const classNames = ['label'];
+
+    if(isTextBox){
```
alexyork: We do if-statements with spaces everywhere else in the codebase, like this: `if (isTextBox) {`. I will fix, amend and push again.

PR #49, opened 2017-12-18
New alert styling against new sketches, full width

jimmythesaint82:


src/components/molecules/Alert/Alert.jsx
```diff
@@ -12,21 +12,24 @@ function classNames(kind) {
 /**
  * Status: *finished*.
  */
-const Alert = ({ kind = "positive", headingText, bodyHtml, links = [] }) =>
+const Alert = ({kind = "positive", headingText, bodyHtml, links = []}) =>
```
alexyork: When we use destructuring in every other place in the codebase, we put spaces in, so we should do that here also. Can we ensure we have an ESLint rule for this too?

src/components/molecules/Alert/Alert.pcss
```diff
@@ -2,27 +2,35 @@
     border-left: none;
     border-right: none;
     margin-bottom: 1rem;
-    padding: 1rem;
+    padding: 1rem 1rem 1rem 1rem;
```
alexyork: This was changed from `padding: 1rem;` to `padding: 1rem 1rem 1rem 1rem;` but I think we should prefer the shorthand syntax.

src/components/molecules/Alert/Alert.pcss
```diff
@@ -2,27 +2,35 @@
...
     @media screen and (min-width: 60em) {
         border: 2px solid var(--dark-green);
-        border-radius: 5px;
         display: flex;
         margin: 1rem auto;
-        width: var(--width-small);
+        width: 100%;
+    }
+
+    &__text-container{
```
alexyork: Space before { !

src/components/molecules/Alert/Alert.jsx
```diff
@@ -12,21 +12,24 @@ function classNames(kind) {
...
+                                <a className="list__link" href={link.url} target="_self">{link.text}</a>
+                            </li>
+                        )}
+                    </ul> : null}
+            </div>
         </div>
+
     </div>;
 
 export default Alert;
```
alexyork: We may use an Alert component in various places on the styleguide website (such as the Download page, and maybe others). Those places must also be updated with the new HTML structure.

PR #52, opened 2018-01-02
Updated alert to default to small, with option for full width

jimmythesaint82:


src/components/molecules/Alert/Alert.jsx
```diff
@@ -3,17 +3,17 @@ import React from 'react';
 
 import Heading from '../../atoms/Heading/Heading';
 
-function classNames(kind) {
+function classNames(kind, size) {
     const classNames = ['alert'];
-    classNames.push(`alert--${kind}`);
+    classNames.push(`alert--${kind} ${size === 'large' ? 'alert--large' : ''}`);
```
alexyork: We tend to do it like this in other places:

```
classNames.push(`alert--${kind}`);
if (size === 'large') {
    classNames.push(`alert--large`);
}
```

i.e. 1 `.push()` per variable. It also would avoid the trailing space at the end of the class name.

src/components/molecules/Alert/Alert.jsx
```diff
@@ -3,17 +3,17 @@ import React from 'react';
...
     return classNames.join(' ');
 }
 
 /**
  * Status: *finished*.
  */
-const Alert = ({ kind = "positive", bodyHtml, links = [] }) =>
-    <div className={classNames(kind)}>
+const Alert = ({ kind = "positive", size = "small", bodyHtml, links = [] }) =>
+    <div className={classNames(kind,size)}>
```
alexyork: Space between parameters in JS, always: `kind,size` -> `kind, size`

src/components/molecules/Alert/Alert.jsx
```diff
@@ -3,17 +3,17 @@ import React from 'react';
...
+    classNames.push(`alert--${kind} ${size === 'large' ? 'alert--large' : ''}`);
     return classNames.join(' ');
 }
 
 /**
  * Status: *finished*.
  */
-const Alert = ({ kind = "positive", bodyHtml, links = [] }) =>
-    <div className={classNames(kind)}>
+const Alert = ({ kind = "positive", size = "small", bodyHtml, links = [] }) =>
```
alexyork: The ` = "small"` has no purpose, and can be removed.

alexyork:
screen shot 2018-01-02 at 13 59 37

Can you also check with Ulrikke if this is correct?

PR #57, opened 2018-01-10
NumberBoxes: replace horizontal scrolling with grid

mhd999:
The changes here to prevent scrolling on `NumbeBoxes` component and instead use grid.

The changes has been tested on Firefox and Chrome, not sure what version of IE we are supporting but since all the styling not working on IE 11 that i have i skipped testing on IE.

### DIFF
--- a/src/components/atoms/NumberBoxes/NumberBoxes.pcss
+++ b/src/components/atoms/NumberBoxes/NumberBoxes.pcss
@@ -2,7 +2,6 @@
     background-color: var(--grey);
     cursor: all-scroll;
     display: flex;
-    flex-wrap: nowrap;
     list-style: none;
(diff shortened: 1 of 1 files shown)


src/components/atoms/NumberBoxes/NumberBoxes.pcss
```diff
@@ -2,63 +2,61 @@
...
-            border-top: 2px solid var(--light-grey);
-        }
-    }
 }
 
 .number-box {
     background-color: var(--grey);
+    border-top: 4px solid var(--white);
+    border-right: 4px solid var(--white);
     border-left: 4px solid var(--white);
```
mhd999: Not sure if we can do line 23, 24 and 25 in one line.

PR #70, opened 2018-01-31
add Helvetica font

mhd999:


cathrholm:
Needs approvement from designer before merge.

PR #73, opened 2018-02-09
Campaign component v1

alexyork:
We anticipate that we will come back to this component and adjust the background image size, once we speak with UX. The task can be reviewed, tested and deployed as it is now, though.


alexyork:
I fixed the alt-text, and will now merge.

PR #75, opened 2018-02-12
add hightlight to step-by-step component

mkmarek:


cathrholm:
We will leave this PR open to clearify some tabbing and styling with designer before we merge.

src/components/atoms/StepByStep/StepByStep.interactive.html
```diff
@@ -0,0 +1,29 @@
...
+        <div tabindex="0" class="step-by-step__circle step-by-step__circle--interactive step-by-step__circle--highlighted">
+            <span class="step-by-step__number">1</span>
+        </div>
+        <div class="step-by-step__text">
+            <h2 class="step-by-step__heading heading heading--level-2">Lorem</h2>
+            <div class="step-by-step__description">This is a short descriptive text. This is a short descriptive text</div>
+        </div>
+    </li>
+    <li class="step-by-step__step">
+        <div tabindex="0" class="step-by-step__circle step-by-step__circle--interactive">
```
cathrholm: Due to accessibility reasons, maybe this should be a button instead of a div? Since it can be tabbed to, we would expect it to trigger a response of some sort when a user clicks the element.

PR #77, opened 2018-02-14
consolidate radio lists

mkmarek:


src/components/molecules/RadioButtonList/RadioButtonList.jsx
```diff
@@ -0,0 +1,27 @@
+import _ from 'lodash';
```
cathrholm: Remove unused import

src/components/molecules/RadioButtonList/RadioButtonList.jsx
```diff
@@ -0,0 +1,27 @@
...
+
+import RadioButtonWithLabel from './RadioButtonWithLabel';
+
+const getClassName = (type) =>
+    classnames('radio-button-list', { 'radio-button-list--horizontal' : type === 'horizontal' });
+
+/**
+ * Status: *in progress*.
+ *
+ * Waiting on the Photoshop files to get the exact checkmark SVG.
```
cathrholm: Is the Status "in progress" still correct? and are we waiting on photoshop-files?

src/components/molecules/RadioButtonList/RadioButtonWithLabel.jsx
```diff
@@ -3,14 +3,14 @@ import React from 'react';
```
cathrholm: I guess this was here from before, but we can remove the unused import here as well

PR #89, opened 2018-03-08
shadow box

mkmarek:


src/components/atoms/ShadowBox/ShadowBox.pcss
```diff
@@ -0,0 +1,12 @@
+.shadow-box {
+    background-color: var(--white);
+    box-sizing: border-box;
+    box-shadow: 0px 4px 15px 0px rgba(0,0,0,0.22);
+
+    margin: 0.625rem;
+    padding: 2.188rem;
+
+    &__container {
```
cathrholm: This css-class is not used in the example.. should we add it to show the consumers of the styleguide that it's a part of the component?

PR #98, opened 2018-03-14
disabled state for textbox and textarea

mkmarek:


src/components/molecules/TextBoxWithLabel/TextBoxWithLabel.jsx
```diff
@@ -9,10 +9,10 @@ import Label from '../../atoms/Label/Label';
  * the user to give correct input.
  *
 **/
-const TextBoxWithLabel = ({ labelText, type, placeholder, errorMessage}) =>
+const TextBoxWithLabel = ({ labelText, type, placeholder, errorMessage, disabled}) =>
```
cathrholm: I can see that this was there before your changes, but add a space after 'disabled'

PR #109, opened 2018-03-19
RecommendedProducts component

cathrholm:


src/components/molecules/RecommendedProducts/RecommendedProducts.jsx
```diff
@@ -0,0 +1,37 @@
...
+
+
+const RecommendedProducts = ({ mainHeading, products, image, heading, text, buttonText }) =>
+    <div className="container container--large">
+        <div className="recommended-products">
+            <Heading level="2" text={mainHeading} className="recommended-products__heading"/>
+            <div className="recommended-products__products">
+                {products.map((product, i) =>
+                    <div className="recommended-products__item-container">
+                        <a key={i} className="recommended-products__item recommended-products__product" href={product.url}>
```
alexyork: `key` needs to be moved up the DOM one level to the root element of the loop, in order to avoid a React warning.

src/components/molecules/RecommendedProducts/RecommendedProducts.jsx
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+
+import Heading from '../../atoms/Heading/Heading';
+import Button from '../../atoms/Button/Button';
+
+
+const RecommendedProducts = ({ mainHeading, products, image, heading, text, buttonText }) =>
+    <div className="container container--large">
```
alexyork: Can this be removed, either by setting the relevant styles on `.recommended-products` or by moving the `.container` classes down onto the `.recommended-products` element? (First one is preferred, I think).

cathrholm: Yes, offcourse. It's done.

src/components/molecules/RecommendedProducts/RecommendedProducts.jsx
cathrholm: Forgot to move this when I added the div around the link. But it's done now.

cathrholm:
Rebasing and bumping version is done right before merging to master, to avoid that you have to change the version number if someone else had committed a change and a new version before you.

PR #117, opened 2018-03-26
Link with icon

Hoferer:


magnusindregard:
image

Needs a white icon on :focus

PR #137, opened 2018-05-30
Add examples for molecules

richiewenn:


alexyork:
# Code formatting

I like the way you format code, but some parts are inconsistent with the existing codebase, and in a codebase where many different devs from different teams will contribute, inconsistency will lead to mess over time, so I feel it is important to get this right now. I admit that ESLint rules and other things like that are lacking in this repo, so this could be an opportunity to fix those things, in line with the comments below.

## Issue 1

The rest of the code base formats like this:

```
{someBool
    ? <div>1</div>
    : <div>2</div>}
```

A lot of your new code formats like this:

```
{someBool
    ? <div>1</div>
    : <div>2</div>
}
```

Note the position of the last `}`. I actually have no strong preference here but my gut feeling is that the first snippet is already heavily in use. I could be wrong about this. If I am not wrong, we need to pick one and stick to it.

## Issue 2

The rest of the code base formats like this:

```
<Component
    prop1="foo"
    prop2="bar" />
```

The examples format like this:

```
<Component
    prop1='foo'
    prop2='bar'
/>
```

The two differences here are that in JSX syntax, we have used double quotes, not single quotes. Also, the trailing `/>` is normally on the same line as the last prop, not a new line. The first snippet code style is already heavily in use in the rest of the codebase, and double quotes are what the official React website uses.

# Functionality

## Full-width components

There was a `componentUtil.isFullWidthComponent()` helper function (still there, now not in use) which would render some component examples full-width. It wasn't the best/cleanest way of doing it, but it worked. It is particularly important when looking at components such as PageHeader, PageFooter and FullWidthImage - which need to be seen in full-screen mode to be fully understood. However this is solved, it should be solved!

## The `warning` library and `deprecated` function

Whilst I think this may be the way we solve deprecating things in the future, _it isn't needed now_ and therefore I am include to say "get rid of it".

A crucial part of our "way of working" is to avoid premature optimizations, avoid adding things before we need them, and writing the _minimum amount of code to solve the issue at hand right now_. With those in mind, I feel we should remove this, and we re-evaluate what to so when/if other channels start to use things more and more, and refactor existing stuff.

## Inline `<code>` tags

In various places, such as the component "docs", we have an inline `<code>` tag. You swapped code-formatting library, it seems, and since then the inline `<code>` tags now have incorrect styling.

PR #158, opened 2018-06-07
Links now hide their internal class names

alexyork:


richiewenn:
I'm just thinking about the `classNames` function. Maybe you should use the `classnames` library, since it's used in many places for doing stuff like this.

```
classnames('link', {
    [`link--with-icon link--${icon}`]: icon,
    [`link--icon-${iconPosition}`]: icon && iconPosition,
    [className]: className,
})
```

alexyork:
Yeah, I actually used that at the start but couldn't think of a clean way to add the classes with that library, since they were not simple "add this class" rules. You managed to do it concisely, though, so I will use your code!

PR #168, opened 2018-06-08
Added StepIndicator component

joakikr:
Some things to point out:

* Sorry for the inline style tag within the component, but it was necessary in order to dynamically compute widths for each step item.


src/client/examples/atoms/StepIndicator/StepIndicator.js
```diff
@@ -0,0 +1,22 @@
...
+    display: 'flex',
+    alignItems: 'center',
+    justifyContent: 'center',
+    height: '60px'
+};
+
+const StepIndicators = () => (
+    <div>
+        <div style={wrapperStyle}>
+            <StepIndicator index={0} numberOfSteps={3} />
```
asmundth: Can these 2 numbers (0,3) be read from a config file?

src/client/examples/atoms/StepIndicator/StepIndicator.js
```diff
@@ -0,0 +1,22 @@
...
+        </div>
+        <div style={wrapperStyle}>
+            <StepIndicator index={2} numberOfSteps={6} />
```
asmundth: Can these 2 numbers (0,3) be read from a config file?

src/client/examples/atoms/StepIndicator/StepIndicator.js
joakikr: Sure can. They are just props, so the parent component may retrieve them however they like, including a `config.js` file. This is just an example, so I think hardcoded values is OK here.

src/client/examples/atoms/StepIndicator/StepIndicator.js
joakikr: Sure can. They are just props, so the parent component may retrieve them however they like, including a `config.js` file. This is just an example, so I think hardcoded values is OK here.

alexyork:
I didn't understand why anything has to be dynamic/calculated here.

This is just an `<ul>` with some `<li>` inside. With something like `display: inline-block;` it will only be as wide as its contents, and then just center-align the whole thing.

The line going through the back of the steps can be done with a `:after` pseudo-selector.

Have I missed something? The implementation here looks overly complicated...

alexyork:
I pushed what I believe is an (almost finished) simpler alternative.

it-vegard:
I merged the two style-examples. The proposal from @alexyork didn't look quite like it should (horizontal line expanded to the edge of the container, fonts were wrong, and a couple of other things. Incorporated your ideas into the previous styling instead, and tried cleaning it up a bit.

The calculating-part was already obsolete after my commit from last week, so I removed the comment about that.

PR #176, opened 2018-06-12
Overhaul of tabs components

alexyork:


package.json
```diff
@@ -62,7 +62,6 @@
     "postcss-nested": "^2.1.2",
     "postcss-replace": "^1.0.3",
     "postcss-url": "^7.3.0",
-    "postcss-for": "^2.1.1",
```
cathrholm: Not that I know what this plugin actually does, but I just wanted to point out that it is deleted.

cathrholm:
From https://www.w3.org/TR/wai-aria-practices/#tabpanel:
- The element that serves as the container for the set of tabs has role tablist.
- Each element that serves as a tab has role tab and is contained within the element with role tablist.
- Each element that contains the content panel for a tab has role tabpanel.
- Each element with role tab has the property aria-controls referring to its associated tabpanel element.
- The active tab element has the state aria-selected set to true and all other tab elements have it set to false.
- Each element with role tabpanel has the property aria-labelledby referring to its associated tab element.

component-lib/src/molecules/Tabs/Tabs.jsx
```diff
@@ -1,27 +1,46 @@
...
-            React.cloneElement(element, { key: index, index, isSelected: index === selectedIndex }))}
+            React.cloneElement(element, {
+                key: index,
+                index,
+                uniqueId,
+                isSelected: (index === selectedIndex)
+            }))}
+    </div>;
+
+Tabs.TabElement = ({ index, uniqueId, isSelected, children }) =>
```
cathrholm: Can we change name from "TabElement" to "TabPanel"? I think it's more descriptive that way.

component-lib/src/molecules/FunkyTabs/FunkyTabs.jsx
```diff
@@ -0,0 +1,54 @@
...
+        {children.map((element, index) =>
+            React.cloneElement(element, {
+                key: index,
+                index,
+                uniqueId,
+                isSelected: (index === selectedIndex)
+            }))}
+    </div>;
+
+FunkyTabs.TabElement = ({ index, uniqueId, isSelected, children }) =>
```
cathrholm: Can we change name from "TabElement" to "TabPanel"? I think it's more descriptive that way.

package.json
alexyork: Jepp, not needed anymore.

component-lib/src/molecules/Tabs/Tabs.jsx
alexyork: Yes We Can! (tm)

component-lib/src/molecules/FunkyTabs/FunkyTabs.jsx
alexyork: Done

component-lib/src/molecules/Tabs/Tabs.jsx
```diff
@@ -1,27 +1,51 @@
...
+                    className={classnames(
+                        'tabs__list-item',
+                        { 'tabs__list-item--selected': (selectedIndex === index) })}
+                    id={`${uniqueId}-tab-${index}`}
+                    role="tab"
+                    aria-controls={`${uniqueId}-panel-${index}`}
+                    aria-selected={selectedIndex === index}>
+                    <button
+                        onClick={() => onSelect(index)}
+                        tabIndex="0"
```
cathrholm: Don't think we need tabIndex when we use a button element?

PR #196, opened 2018-06-28
LabelWithIcon component

richiewenn:


richiewenn:
Oh, yeah, it should be possible to modify Tooltip and use it. I'm closing this issue.

PR #199, opened 2018-06-28
Add login/out actions to PageHeader and PageMenu

richiewenn:


ropimo:
> ... loginLink and logoutLink = "#" ... get the correct styling ...
yep, it's easy but it smells more of a patch than a clean fix

PR #209, opened 2018-07-24
image in accordion heading

mkmarek:


alexyork:
Let's not over-complicate nice, simple components in this way. Closing the PR in the name of clean UI's and simplicity!

PR #217, opened 2018-08-03
Feature/remove sprites

marenpg:


marenpg:
There was some trobles after the merge which I will now correct. I'll post here when it is fixed.

marenpg:
Should be ok now :)

alexyork:
Merged in another branch.

PR #220, opened 2018-08-07
changed svg and image tags to use a new svg icon component

marenpg:


component-lib/src/atoms/List/List.pcss
```diff
@@ -102,7 +102,7 @@
             padding: 1px;
 
             .list__link-icon {
-                svg {
+                svg{
```
marenpg: This should have a space to follow the correct format, but this will be changed in a feature to update the use of icons in lists and links.

component-lib/src/molecules/FactBox/FactBox.jsx
```diff
@@ -19,10 +20,10 @@ const FactTwoColumnContent = ({ iconSrc, title, centerIcon, imgSrc, imgAlt, chil
     </React.Fragment>
 )
 
-const FactContent = ({ children, iconSrc, title, underline }) => (
+const FactContent = ({ children, iconName, title, underline }) => (
     <React.Fragment>
-        {iconSrc && (
-            <img className="fact-box__icon" src={iconSrc} alt={title} />
+        {iconName && (
+            <SvgIcon className="fact-box__icon" iconName={iconName} color="black" />
```
marenpg: In this component (and some others) I was wondering if we should make it possible to set icon color as well.

PR #236, opened 2018-09-05
Added a scaled version of StepIndicator at 37,5em breakpoint

joakikr:


it-vegard:
I would like to try out the vw unit for anything above the minimum size here, but that can be done in a later version.

PR #247, opened 2018-09-14
[WIP]  move allicons to sep files and use require to use them

ev1stensberg:
Rollup isn't cooperating well with our current project setup. We'll need to require the svg files, as we cannot fetch files to our public folder. If we fetch from `public`, then the only thing working with svg's would be the server, not when people are trying to import our components.


alexyork:
We have solved this in another way now ;-)

ev1stensberg:
Sweet! Mind linking to it?

PR #266, opened 2018-10-08
Updated StepIndicator atom with new design

joakikr:


component-lib/src/atoms/StepIndicator/StepIndicator.pcss
```diff
@@ -1,28 +1,79 @@
...
         display: flex;
         justify-content: center;
         align-items: center;
-        background-color: white;
-        border: var(--step-indicator-border) solid var(--grey-darker);
+        background-color: var(--dark-grey);
+        border: var(--step-indicator-border) solid var(--dark-grey);
         border-radius: 50%;
-        color: var(--grey-darker);
+        color: var(--dark-grey-2);
```
it-vegard: This text color calculates to 1.93 contrast against --dark-grey. Minimum requirement is 3.0 for WCAG 2.0 AA, and 4.5 for AAA level conformance.

One option is to switch to white color on a bit darker background, like #949494 (instead of #A0A0A0), but we could probably go even a bit darker background like that.

Alternatively, we could darken only the text color, to e.g. #525151, but I don't think that looks good.

component-lib/src/atoms/StepIndicator/StepIndicator.pcss
```diff
@@ -1,28 +1,79 @@
...
+    --step-indicator-icon-size: 0.6rem;
 }
 
 .step-indicator {
     display: flex;
     justify-content: center;
 
+    &__label {
+        font-size: 0.75rem;
+        color: var(--dark-grey);
```
it-vegard: This color does not validate color contrast against white background (2.61).
Should switch to --dark-grey-2, as for active, which has a contrast of 5.03.

component-lib/src/atoms/StepIndicator/StepIndicator.pcss
it-vegard: Also adding that the requirements in WCAG 2.1 is 4.5 and 7.0. This will be included in the law possibly as early as next summer, so we should aim higher than a contrast of 3.0 for future-proofing reasons.

PR #291, opened 2018-12-12
Using html instead of text in accordion

chrinaldini:


mhadaily:
We can pass down a React component as children so we don't need this.

PR #294, opened 2019-01-02
Updated packages to solve network alerts

marenpg:


marenpg:
Updated packages based on some: https://github.com/TeliaSoneraNorge/styleguide/network/alerts

PR #308, opened 2019-02-04
Add vertical list display option for USPs

AmundLochen:


alexyork:
How will anyone know to use this, if there is no example for those browsing the Styleguide website, and the React component accepts no props to implement it? :-)

AmundLochen:
Now they should be able to find out about this functionality through careful study of the styleguide instructions instead of having to rely on intuition and guesswork. :)

alexyork:
Good job. I'll get this onto master. You also win the prize for the funniest GitHub comment of the week.

PR #309, opened 2019-02-04
added featureboxees to styleguide

AnmerSeif:


component-lib/src/organisms/FeatureBoxes/FeatureBoxes.pcss
```diff
@@ -0,0 +1,46 @@
...
+        margin:0 2.25rem;
+        margin-bottom:2.8125rem;
+        &__icon{
+            text-align: center;
+            margin-bottom:1.1875rem;
+            svg{
+                width:3rem;
+                height: 3rem;
+            }
+        }
```
alexyork: Check other CSS files for formatting. In all other places, we have done this:
`.feature-box {` instead of `.feature-box{`.

And this:
`width: 3rem;` instead of `width:3rem;`

Let's be consistent with the rest of the codebase!

src/client/examples/organisms/FeatureBoxes/Default.js
```diff
@@ -0,0 +1,58 @@
...
+
+const FeatureBoxesDemo = () => (
+    <FeatureBoxes>
+        <FeatureBox
+            id="featurebox-1"
+            iconSvg={<SvgIcon iconName="ico_desktop" color="purple" />}
+            heading="selvbetjening"
+            url="http://telia.no"
+            text="Ønsker du å gi dine ansatte større frihet, og samtidig sikre .. "
+            />
```
alexyork: Where have we added `/>` in all other examples?

alexyork:
Ok, the code now looks decent - but I can't actually _see_ the FeatureBoxes anywhere, so I can't test that it looks good. They do not show up when I run the styleguide website. It should appear as an organism, and the examples file you wrote should be rendered. Am I doing something wrong?

PR #315, opened 2019-02-15
Removed pointer-events from accordion icon

AnmerSeif:


component-lib/src/molecules/AccordionList/AccordionList.pcss
```diff
@@ -2,19 +2,18 @@
...
 
-* ~ .accordion {
+*~.accordion {
     border-top: var(--accordion-border);
 }
 
-p + .accordion,
-p + .accordion-list {
+p+.accordion,
+p+.accordion-list {
```
marenpg: I think the spaces should be kept.

alexyork: Me too. It looks like Anmer has some tool in his IDE which is doing auto-formatting. That is going to put lots of changes in each commit which distracts from the actual code being changed.

@AnmerSeif Can you fix?

One habit I highly recommend is to go through each file with a fine tooth-comb at all lines changed. Only stage & commit the lines that are related to the feature you are implementing.

PR #330, opened 2019-02-28
Added ability to choose device in DisplayFunctionalityBlock

AnmerSeif:


marenpg:
It looks good. I moved the image url to the outer website so that people importing the react component can add their own image as website-images are not available in the component-lib. The ideal solution would be to bundle the images to the component, however, I tried this using [rollup-plugin-image](https://github.com/rollup/rollup-plugin-image) but it gave errors.

PR #332, opened 2019-03-01
Replace old quote-style with new styling

AmundLochen:


component-lib/src/atoms/Quote/Quote.jsx
```diff
@@ -5,21 +5,25 @@ import classnames from 'classnames';
 /**
  * Status: *finished*
  */
-const Quote = ({ children, className, inline = false, source, ...rest }) => (
+const Quote = ({ children, className, inline = false, variant = false, source, ...rest }) => (
```
marenpg: Instead of the variant being a boolean, maybe it can be a type with limit on valid input? That would be a bit more descriptive that just "variant" which doesn't say much about what variant it is.

PR #354, opened 2019-03-21
Added active underline to menu heading

AnmerSeif:


component-lib/src/molecules/Menu/Menu.jsx
```diff
@@ -43,7 +43,9 @@ export default class Menu extends React.Component {
                                 return (
                                     <li key={menuLink.heading+index}>
                                         <a href={`${menuLink.heading}`}   
-                                                  className="menu__heading-item link">
+                                                className={ classnames('menu__heading-item link',{
```
mkmarek: Shouldn't be in state. Should be a prop instead so channels can set their own preferred heading to be set.

Or it could be a combination that the initial value would be taken from a prop and then it's stored in state

PR #367, opened 2019-03-26
added babel plugin to avoid having export/import ES6 syntax in the build files

smidhonza:
without it in `dist/index.js`
```
// Atoms
export { default as AlertText } from './atoms/AlertText';
export { default as Box } from './atoms/Box';
...
```

with the plugin
```
"use strict";

Object.defineProperty(exports, "__esModule", {
  value: true
});
Object.defineProperty(exports, "AlertText", {
  enumerable: true,
  get: function get() {
    return _AlertText.default;
  }
});
...
```

### DIFF
--- a/component-lib/babel.config.js
+++ b/component-lib/babel.config.js
@@ -12,6 +12,7 @@ module.exports = {
         ['@babel/plugin-proposal-class-properties', { loose: true }],
         '@babel/plugin-proposal-object-rest-spread',
         '@babel/plugin-transform-object-assign',
+        '@babel/plugin-transform-modules-commonjs',
         'lodash'
     ],
     env: {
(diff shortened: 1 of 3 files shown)


mkmarek:
Exporting in ES6 is required for webpack's tree shaking https://webpack.js.org/guides/tree-shaking/.
Sorry, cannot merge this. I noticed there is a problem in ReCss while using Jest and found a remedy for that. You can have a look at: https://github.com/TeliaSoneraNorge/re-css/pull/159/commits/7320b5e59f02bd7b923e54d217ef4a34684b6412

smidhonza:
@mkmarek ok I used your fix in recss and that worked fine, thanks! closing this

PR #373, opened 2019-03-28
New new menu

marenpg:


marenpg:
Will fix opening of menu in IE11, and verify that it works in safari on Monday

PR #375, opened 2019-03-29
Decrease heading size slightly

AmundLochen:


marenpg:
Not sure why this was requested, or by whom, but if it should be merged all channels should agree?

PR #376, opened 2019-04-01
Added CardsBlock

AnmerSeif:


component-lib/src/molecules/Cards/Card.jsx
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * Status: *finished*.
+ */
+const Card = ({ children, className, isGrey, img, title, link, ...rest }) => (
+    <div className="card__wrapper">
+        <a className={ classnames('card link', {
+            'card--background-grey' : isGrey
+        })} href={link}>
+            <img className="card__image" src={img} />
```
marenpg: We should have alt-tag on images

component-lib/src/molecules/Cards/Card.jsx
```diff
@@ -0,0 +1,29 @@
...
+                {children}
+            </div>
+        </a>
+    </div>
+);
+
+Card.propTypes = {
+    title: PropTypes.string,
+    isGrey: PropTypes.bool,
+    img: PropTypes.string
```
mkmarek: There are some missing proptypes here

PR #388, opened 2019-04-12
replaced defaultValue with value prop in <select /> element in <DropDownListWithLabel />

smidhonza:
<DropDownListWithLabel />  is controlled component so it shouldn't have `defaultValue` prop but `value` (https://reactjs.org/docs/forms.html#controlled-components). With current `defaultValue` the select options won't update when `value` prop has changed


smidhonza:
no comments from others so I guess I'll merge it. It is similar change I've done before for `< RadioButtonWithLabel  />` in https://github.com/TeliaSoneraNorge/styleguide/pull/382 so it _shouldn't_ break anything

PR #394, opened 2019-05-06
Subscription box: Moved all hardcoded text to props

marenpg:


component-lib/src/molecules/Subscription/Subscription.jsx
```diff
@@ -34,103 +45,55 @@ const Subscription = ({ isShowingFeatures, isStandalone, isBroadband, color, siz
...
-                        <SvgIcon className="subscription__highlighted-feature-icon" iconName={highlightedFeature.iconName} role="presentation" alt="" />
-                        <span className="subscription__highlighted-feature-text">{highlightedFeature.name}</span>
-                    </div> : null}
-                {specialMessageText ?
-                    <strong className="special-message">{specialMessageText}</strong> : null}
-                {isBroadband &&
-                    <PriceTable productListWithPrice={productList} totalTextWithPrice={totalCalculation} />}
-                <button className="button button--primary">Gå videre</button>
-            </section> : null}
+        {(features) ?
```
it-vegard: We currently have features that are only shown while expanded:
https://www.telia.no/mobilabonnement/mobilabonnement-for-barn
Changing from `(isExpanded || isShowingFeatures)` to `(features)` will mean that is no longer supported.

Also, the current implementation will mean the "Gå videre" button will not be shown for Subscriptions lacking features.

In short, I think we still need some form of flag to support features that are not shown when not expanded, and at least the "Gå videre" button must be added even if there are no features.

marenpg: I added back the `isShowingFeatures` that will determin if features are shows on collapsed boxes. To fix a bug that features are not rendered on subscriptions that start off collapsed and do not show features I will keep the `(features)` instead of `(isExpanded || isShowingFeatures)`. 
I updated the feature-button so it takes a button component (more room for customization) and to only display the button, it can be added in the feature component (as it actually lives inside the feature section). :)

PR #396, opened 2019-05-23
negative Button kind

ropimo:


src/components/examples/atoms/Button/Buttons.js
```diff
@@ -17,6 +17,7 @@ const Buttons = () => (
         <Button text="Confirm" />
         <Button text="Cancel button" kind="cancel" />
         <Button text="Inverted button" kind="inverted" />
+        <Button text="Negative button" kind={Button.kinds.negative} />
```
marenpg: Should we update the rest to use `Button.kinds` as well maybe?

ropimo: I didn't want to impose 'my style' over the other. Both approaches work fine.

PR #397, opened 2019-05-23
Added the missing dependency postcss-url

BjornEvenEgge:


package.json
```diff
@@ -62,6 +62,7 @@
     "object-fit-images": "3.2.4",
     "parcel-bundler": "1.12.3",
     "picturefill": "3.0.3",
+    "postcss-url": "^8.0.0",
```
it-vegard: Seems like something that should be in `devDependencies`?

PR #398, opened 2019-05-23
Radiobutton tweaks

BjornEvenEgge:
Minor adjustments to the styling of radio buttons

### DIFF
--- a/component-lib/src/molecules/RadioButtonList/RadioButtonWithLabel.jsx
+++ b/component-lib/src/molecules/RadioButtonList/RadioButtonWithLabel.jsx
@@ -4,7 +4,7 @@ import PropTypes from 'prop-types';
 import Label from '../../atoms/Label/Label';
 
 const RadioButtonWithLabel = ({ label, checked, disabled, name, value, hasRichContent, onChange, children }) =>
-    <Label className="label--thin radio-button-with-label">
+    <Label className="radio-button-with-label">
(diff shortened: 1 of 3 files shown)


component-lib/src/molecules/RadioButtonList/RadioButtonWithLabel.jsx
```diff
@@ -4,7 +4,7 @@ import PropTypes from 'prop-types';
 import Label from '../../atoms/Label/Label';
 
 const RadioButtonWithLabel = ({ label, checked, disabled, name, value, hasRichContent, onChange, children }) =>
-    <Label className="radio-button-with-label">
+    <Label className="label--thin radio-button-with-label">
```
marenpg: This can maybe be optional, so you can choose thin label or not? Or has it been decided that all radio buttons should have thin labels in telia?

BjornEvenEgge: This style should be the default setting for radiobuttons, but I'm not sure whats the best practice for adding it? Is it safest to make it optional?

alexyork: If this is the default styling for radios, we can simply add it to the styles of the radio - no need for a class name, or to touch the label component. I'll do it.

PR #406, opened 2019-06-13
created disabled version of dropdownlist

georgpal:


component-lib/src/molecules/DropDownListWithLabel/DropDownListWithLabel.pcss
```diff
@@ -29,6 +29,11 @@
             width: 50%;
         }
 
+        &--disabled {
+            background: transparent url('{{assetPath}}/icons/ico_dropArrow_grey.svg') no-repeat right 10px center;
+            background-size: 15px 15px;
+        }
+
         &::-ms-expand {
```
marenpg: Here you can change only `background-image` instead of the whole background property :)

PR #418, opened 2019-06-27
Feature/add eslint and prettier

it-vegard:
- Adding Prettier and ESLint to the Styleguide
- Running Prettier and ESLint on only files being commited, using Husky and Lint-Staged
- (You can also run ESLint manually using `npm run lint`, of course)
- Fix warning from ESLint with non-declared React version


ropimo:
It's looking good now.

(on another page, the _force-push_ sounds quite rapist... maybe a gentlemen's _merge from master_ will proceed with mutual consent)

PR #428, opened 2019-08-27
npm install before npm run build

ropimo:
not sure how the concourse pipeline tasks and cache work at all. But shouldn't it contain `npm install` somewhere before `npm run build` to handle the new dependencies?


marenpg:
We use this resource type: https://github.com/ymedlop/npm-cache-resource that does the npm install and caches the node_modules so we do not have to install.

PR #429, opened 2019-08-31
separate script to add icons

ropimo:
The script search for new icons in a _dropHere_ folder, optimizes them and generates the react code for allIcons.

Also new style of displaying icon in styleguide-web


component-lib/package.json
```diff
@@ -7,15 +7,15 @@
   "jsnext:main": "dist/es/index.js",
   "module": "dist/es/index.js",
   "scripts": {
-    "build": "run-p build:commonjs build:es build:css build:icons && run-s build:copy-files",
-    "build:allicons-react": "node generate-AllIcons-component.js && babel ./assets/allicons/AllIcons.jsx.generated --out-file ./assets/allicons/AllIcons.js",
+    "icons": "run-s icons:addIcons icons:svg-join icons:allicons-react",
```
marenpg: Maybe use kebab-case on the name for consistency so that it is `icons:add-icons`?

marenpg:
It looks good, but I think I miss having the unoptimized icons available somewhere. Sometimes I've had to change something in the icon, and that's pretty hard in the optimized versions. F.ex. when the new login button was made, the designers wanted a slimmer border in the login icon and I could just fix that directly in the svg instead of them having to make a new one. Can we maybe keep a folder next to icons called optimized like we had for that?

PR #431, opened 2019-09-10
Use LinkTemplate prop for Login and MinSide URL links

ropimo:
replace hardcoded **<a>** tag by configurable **LinkTemplate** prop for _Login_ and _Min Side_ links

### DIFF
--- a/component-lib/package.json
+++ b/component-lib/package.json
@@ -19,6 +19,7 @@
     "clean": "rimraf dist",
     "deploy:prod": "npm run build && npm publish dist",
     "deploy:beta": "npm run build && npm publish dist --tag beta",
+    "dev:es": "cross-env BABEL_ENV=es babel src --out-dir dist/es --watch",
     "lint": "eslint --ext .jsx,.js ./src",
     "prebuild": "run-s clean",
     "publish:local-test": "npm run build && cd dist && npm pack",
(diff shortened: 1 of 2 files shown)


component-lib/package.json
```diff
@@ -19,6 +19,7 @@
     "clean": "rimraf dist",
     "deploy:prod": "npm run build && npm publish dist",
     "deploy:beta": "npm run build && npm publish dist --tag beta",
+    "dev:es": "cross-env BABEL_ENV=es babel src --out-dir dist/es --watch",
```
marenpg: So we can run the component lib locally? :D

component-lib/src/molecules/Menu/MenuContent.jsx
```diff
@@ -62,18 +62,18 @@ const MenuLinkItem = ({ link, LinkTemplate, onToggleSubmenu, isSubmenuOpen, isAc
     </li>
 );
 
-const LoginButton = ({ loginUrl }) => (
-    <a href={loginUrl} className="menu__login-button button button--small">
+const LoginButton = ({ loginUrl, LinkTemplate }) => (
+    <LinkTemplate className={'menu__login-button button button--small'} url={loginUrl}>
```
marenpg: Does this have to be in brackets? Can it just be string?

component-lib/package.json
ropimo: it watches the component-lib for changes and build the ES version. Useful when using `yarn link` to see the _styleguide_ changes in another project

PR #454, opened 2019-10-23
Accessible Funky tabs

mkmarek:


package.json
```diff
@@ -79,7 +79,8 @@
     "@telia/styleguide": "./component-lib/src",
     "component-lib": "./component-lib/src",
     "component-lib-root": "./component-lib",
-    "assets": "./src/assets"
+    "assets": "./src/assets",
+    "react": "./node_modules/react"
```
mkmarek: Parcel was taking react from two different places, since it takes libs from the nearest node_modules folder and we have on in the component-lib folder. (ref: https://parceljs.org/module_resolution.html)

That messed up usage of hooks.

That alias will force it to take it only from single place.

component-lib/src/molecules/FunkyTabs/FunkyTabs.jsx
```diff
@@ -17,53 +19,61 @@ import PropTypes from 'prop-types';
...
+            const previousIndex = (selectedIndex - 1 + children.length) % children.length;
+            const previousElement = children.find((_, index) => index === previousIndex);
+            onSelect(e, previousIndex, previousElement.props.url, previousElement.props.heading);
+        } else if (e.which === 39) {
+            const nextIndex = (selectedIndex + 1) % children.length;
+            const nextElement = children.find((_, index) => index === nextIndex);
+            onSelect(e, nextIndex, nextElement.props.url, nextElement.props.heading);
+        } else if (e.which === 40 && containerRef.current) {
+            containerRef.current.focus();
+        }
```
joakikr: Maybe add a comment beside the number `37, 39 and 40` to make it easier to identity which letter they correspond too.

component-lib/src/molecules/FunkyTabs/FunkyTabs.jsx
```diff
@@ -17,53 +19,61 @@ import PropTypes from 'prop-types';
...
+        tabIndex="-1"
+        ref={containerRef}
         className={classNames(
             'funky-tabs__panel',
             { 'funky-tabs__panel--selected': isSelected })}
         id={`${uniqueId}-panel-${index}`}
         aria-labelledby={`${uniqueId}-tab-${index}`}
-        role="tabpanel">
+        role="tabpanel"
+        hidden={isSelected}>
```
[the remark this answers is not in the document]
mkmarek: Yep, it should be. Fixed.

component-lib/src/molecules/FunkyTabs/FunkyTabs.jsx
mkmarek: I instead extracted those number into consts so it's more descriptive. I also extracted the index calulating and tab getting logic into separate function so it looks prettier.

PR #455, opened 2019-10-24
Dialogs V2

mkmarek:
**Warning breaking change!**

Previously modals were displayed by providing `active` prop to the modal overlay like this:

```jsx
<ModalDialogOverlay active={this.state.confirmDialogActive}>
    ...modal stuff
</ModalDialogOverlay>
```

The modal overlay component has been completely removed and overlay is now part of the `ModalDialog` component.

New component that has been added is `ModalDialogProvider` which should be defined at the application root and provides a container where all dialogs would be instantiated to prevent instantiating dialogs nested way deep in the DOM.

In the PR styleguide puts it in App.js https://github.com/TeliaSoneraNorge/styleguide/pull/455/files#diff-57aa11691094c3295e1f9dcbe8432ba2R9

It works by providing `ref` within its own context to all dialog components under it. The `ModalDialog` component then uses portals to insert their DOM into appropriate container.

Since there is no `active` prop and no `ModalDialogOverlay` dialogs need to be conditionally rendered and there should be only one dialog rendered at a single moment.

```jsx
{this.state.confirmDialogActive &&
    <ModalDialog
        name="confirm-dialog"
        heading="Confirm dialog"
        onSubmit={this.toggleConfirmModalDialog}
        submitText="Confirm"
        onClose={this.toggleConfirmModalDialog}
        closeText="Cancel">
        <p>
            This dialog can be associated with actions where it's necessary to get the user to
            confirm what's about to happen.
    </p>
    </ModalDialog>}
```

Also there are some prop changes:

`onClickButton1` has been renamed to `onSubmit`
`buttonText1` has been renamed to `submitText`
`onClickButton2` has been renamed to `onClose`
`buttonText2` has been renamed to `closeText`

Also this PR defines new component called `FocusTrap`. It's designed to prevent tabbing outside a specific container. This component can be later moved to a common folder of some sort once we need to use it in multiple components.

Dialogs had the option to be rendered as a standalone element on the page. This option is kept with using the `standalone` prop.


component-lib/src/molecules/ModalDialog/FocusTrap.jsx
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+
+const KEY_TAB = 9;
+export const focusableElementsSelector = 'a[href], area[href], input:not([disabled]), select:not([disabled]), textarea:not([disabled]), button:not([disabled]), [tabindex="0"]';
+
+/**
+ * Status: *in progress*.
+ * Category: Notifications
+ *
+ * Prevens focusing outside the container with using TAB keys
```
marenpg: typo

PR #470, opened 2019-11-07
Add QuantityPicker component

alfredbirk:
Add new atom: QuantityPicker. My first Styleguide component, so please tell me if something needs changes.

image
I couldn't get the documentation to work as seen above

### DIFF
--- a/component-lib/src/atoms/QuantityPicker/QuantityPicker.jsx
+++ b/component-lib/src/atoms/QuantityPicker/QuantityPicker.jsx
@@ -2,62 +2,53 @@ import React  from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
 import SvgIcon from '../../../dist/atoms/SvgIcon';
+import QuantityButton from './QuantityButton';
 
 /**
  * Status: *finished*.
  * Category: Buttons
  */
-const QuantityButton = ({ icon, isDisabled, onClick, text }) => (
-    <button
-        className={classnames('button button--small quantity-picker__button', {
-            'button--disabled': isDisabled
-        })}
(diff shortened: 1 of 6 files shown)


component-lib/src/atoms/QuantityPicker/QuantityPicker.jsx
```diff
@@ -0,0 +1,70 @@
...
+const QuantityButton = ({ icon, isDisabled, onClick, text }) => (
+    <button
+        className={classnames('button button--small quantity-picker__button', {
+            'button--disabled': isDisabled
+        })}
+        disabled={isDisabled}
+        onClick={onClick}>
+        <SvgIcon className="quantity-picker__button__icon" iconName={icon} title={text} />
+    </button>
+);
```
joakikr: If you move this out in its own file (say under `QuantityPicker/QuantityButton.jsx`) you will get the documentation working for `QuantityPicker`.

component-lib/src/atoms/QuantityPicker/QuantityPicker.jsx
```diff
@@ -0,0 +1,70 @@
...
+        maxQuantity = Number.MAX_SAFE_INTEGER;
+    }
+
+    return (
+        <div className="quantity-picker">
+            <QuantityButton
+                icon="ico_minus"
+                isDisabled={quantity <= minQuantity}
+                onClick={() => onChangeQuantity(Math.max(quantity - 1, minQuantity))}
+                text="Øk antall"
```
joakikr: Both "Øk antall" and "Reduser antall" should probably be passed as props. This can, however, be their default values.

PR #473, opened 2019-11-14
Fix warnings in QuantityPicker and ToggleButton, change font-size of Quote

alfredbirk:
- Fix warning in ToggleButton component by setting defaultChecked to undefined if the prop is not explicitly set
- Fix warning in QuantityPicker which used to happen when erasing the input field
- Minor change to font-size of text in Quote component so it's more responsive. This is the sizes we use in RapidShop, so thought I would make the same change here.
- Add a name field to the input component of QuantityPicker for improved accessability.

Looks like the linter also automatically makes some formatting changes when pushing.

### DIFF
--- a/component-lib/src/atoms/QuantityPicker/QuantityPicker.jsx
+++ b/component-lib/src/atoms/QuantityPicker/QuantityPicker.jsx
@@ -7,7 +7,7 @@ import QuantityButton from './QuantityButton';
  * Category: Buttons
  */
 
-const QuantityPicker = ({ onChangeQuantity, quantity, minQuantity, maxQuantity, addText, reduceText }) => {
+const QuantityPicker = ({ onChangeQuantity, quantity, minQuantity, maxQuantity, addText, reduceText, inputName }) => {
   if (isNaN(quantity)) {
     quantity = '';
   }
@@ -29,7 +29,7 @@ const QuantityPicker = ({ onChangeQuantity, quantity, minQuantity, maxQuantity, 
             onChangeQuantity(Math.max(Math.min(parseInt(event.target.value), maxQuantity), minQuantity))
           }
           type="number"
-          name="Tast inn antall"
+          name={inputName}
           value={quantity}
         />
       </label>
@@ -50,13 +50,15 @@ QuantityPicker.propTypes = {
   maxQuantity: PropTypes.number,
   addText: PropTypes.string,
   reduceText: PropTypes.string,
+  inputName: PropTypes.string,
 };
 
 QuantityPicker.defaultProps = {
   minQuantity: Number.MIN_SAFE_INTEGER,
   maxQuantity: Number.MAX_SAFE_INTEGER,
   addText: 'Øk antall',
-  reduceText: 'Reuser antall',
+  reduceText: 'Reduser antall',
+  inputName: 'Tast inn antall',
 };
 
 QuantityButton.propTypes = {
--- a/component-lib/src/atoms/QuantityPicker/QuantityPicker.jsx
+++ b/component-lib/src/atoms/QuantityPicker/QuantityPicker.jsx
@@ -30,8 +30,9 @@ const QuantityPicker = ({
         text={reduceText}
       />
 
-      <span className="quantity-picker__input-container">
+      <span className="quantity-picker__container">
         <TextBoxWithLabel
+          className="quantity-picker__input"
           max={maxQuantity}
           min={minQuantity}
           onChange={event =>
--- a/component-lib/src/atoms/QuantityPicker/QuantityPicker.jsx
+++ b/component-lib/src/atoms/QuantityPicker/QuantityPicker.jsx
@@ -25,11 +25,12 @@ const QuantityPicker = ({
     <div className="quantity-picker">
       <QuantityButton
         icon="ico_minus"
-        isDisabled={quantity <= minQuantity || quantity === ''}
+        isDisabled={quantity <= minQuantity || !quantity}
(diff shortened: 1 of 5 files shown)


component-lib/src/atoms/QuantityPicker/QuantityPicker.jsx
```diff
@@ -7,53 +7,63 @@ import QuantityButton from './QuantityButton';
...
 
 QuantityPicker.defaultProps = {
-    minQuantity: Number.MIN_SAFE_INTEGER,
-    maxQuantity: Number.MAX_SAFE_INTEGER,
-    addText: 'Øk antall',
-    reduceText: 'Reuser antall'
+  minQuantity: Number.MIN_SAFE_INTEGER,
+  maxQuantity: Number.MAX_SAFE_INTEGER,
+  addText: 'Øk antall',
+  reduceText: 'Reuser antall',
```
marenpg: Reduser? :)

component-lib/src/atoms/QuantityPicker/QuantityPicker.jsx
```diff
@@ -7,53 +7,63 @@ import QuantityButton from './QuantityButton';
  * Category: Buttons
  */
 
-const QuantityPicker = ({ onChangeQuantity, quantity, minQuantity, maxQuantity, addText, reduceText }) => (
+const QuantityPicker = ({ onChangeQuantity, quantity, minQuantity, maxQuantity, addText, reduceText }) => {
```
marenpg: Can we have a label text (passed as prop) for the input field for accessibility?

component-lib/src/atoms/QuantityPicker/QuantityPicker.jsx
```diff
@@ -7,53 +7,63 @@ import QuantityButton from './QuantityButton';
...
+      <label className="quantity-picker__label">
+        <input
+          className="textbox quantity-picker__input"
+          max={maxQuantity}
+          min={minQuantity}
+          onChange={event =>
+            onChangeQuantity(Math.max(Math.min(parseInt(event.target.value), maxQuantity), minQuantity))
+          }
+          type="number"
+          name="Tast inn antall"
```
marenpg: I don't think we need a name here as we are not referencing the value by name anywhere. If we need it for a form it be a prop. However, as we use react we don't really use the field in that sense.

component-lib/src/atoms/ToggleButton/ToggleButton.jsx
```diff
@@ -11,25 +11,44 @@ import cn from 'classnames';
...
-        <input id={id} className="toggle-button__checkbox" type="checkbox" onChange={onChange} checked={checked} defaultChecked={defaultChecked} />
-        <label htmlFor={id} className="toggle-button__label">{labelText}</label>
-        {descriptionText ? <p className="toggle-button__description">{descriptionText}</p> : null}
-    </div>
+const ToggleButton = ({
+  id,
+  labelText,
+  descriptionText,
+  defaultChecked = undefined,
+  checked,
```
marenpg: What warning is it you get here? When it has no default value it behaves the same for me locally.

marenpg:
For next time: It's better if each of these component changes were separate pull requests do that in case something fails on other sites, like the new quote size, we can easily roll back both commit and version number without it removing anything else. 💯

marenpg:
Also, as the quote change will affect other solutions, the `VERSIONS.md` should note the change I think.

component-lib/src/atoms/ToggleButton/ToggleButton.jsx
alfredbirk: If the `checked` property is used I get "Warning: ToggleButton contains an input of type checkbox with both checked and defaultChecked props. Input elements must be either controlled or uncontrolled (specify either the checked prop, or the defaultChecked prop, but not both)."

marenpg: In rapidshop solution? It seems that the toggle should either be uncontrolled and use `defaultChecked `or controlled and use `checked `with `onChange`. When I tested locally in styleguide with `checked `and `onChange `I don't get this warning about `defaultChecked`.

component-lib/src/atoms/QuantityPicker/QuantityPicker.jsx
```diff
@@ -16,26 +25,26 @@ const QuantityPicker = ({ onChangeQuantity, quantity, minQuantity, maxQuantity,
     <div className="quantity-picker">
       <QuantityButton
         icon="ico_minus"
-        isDisabled={quantity <= minQuantity}
+        isDisabled={quantity <= minQuantity || quantity === ''}
```
marenpg: Can also be written as `!quantity`. :)

component-lib/src/atoms/ToggleButton/ToggleButton.jsx
alfredbirk: I get the same in master branch of styleguide when I test with this for example:
´<ToggleButton id="example-1" labelText="Example with short label text" checked onChange={() => false} />´

component-lib/src/atoms/QuantityPicker/QuantityPicker.pcss
```diff
@@ -4,26 +4,27 @@
...
-            color: var(--white);
-        }
-
-        &::-webkit-inner-spin-button,
-        &::-webkit-outer-spin-button {
-            -webkit-appearance: none;
+    &__input-container {
+        margin: 0 1rem;
+        
+        input {
```
marenpg: Should still give classname here I think rather than styling on the element directly.

PR #511, opened 2020-01-07
Make subscriptions selectable

alfredbirk:


component-lib/src/molecules/Subscription/Subscription.jsx
```diff
@@ -77,6 +83,12 @@ const Subscription = ({
               <div>{additionalInfo.binding}</div>
             </div>
           )}
+          {showFeaturesInTeaser && hasUnlimitedData && (
+            <div className="subscription__feature-info">Fri bruk av data</div>
+          )}
+          {showFeaturesInTeaser && hasMusicFreedom && (
+            <div className="subscription__feature-info">Music Freedom inkludert</div>
+          )}
```
joakikr: Could we rather have the logic for individual features in the parent component, and just have two props here: `showFeaturesInTeaser` and the list `teaserFeatures`. (Or if it's not necessary with the flag, just have the `teaserFeatures` arrray)

Then just `.map()` over the array and display a `<div>...</div>` for each.

PR #520, opened 2020-01-20
Do not print ',-' for price which are not rounded.

mahu9119:


component-lib/src/molecules/Subscription/Subscription.jsx
```diff
@@ -242,4 +242,15 @@ Subscription.propTypes = {
   onClose: PropTypes.func,
 };
 
+const printPrice = price => {
```
mkmarek: Should be probably named `formatPrice`

component-lib/src/molecules/Subscription/Subscription.jsx
```diff
@@ -242,4 +242,15 @@ Subscription.propTypes = {
...
+  let print = false;
```
mkmarek: Should be named `isDecimalNumber` or something more explanatory.

component-lib/src/molecules/Subscription/Subscription.jsx
```diff
@@ -242,4 +242,15 @@ Subscription.propTypes = {
...
+  if (typeof price === 'number') {
+    print = price % 1 === 0;
+  }
+  if (typeof price === 'string') {
+    print = price.indexOf('.') === -1;
```
mkmarek: It's probably better to drop the string comparison completely. Because I could already send in a formatted number like this: `12,4 kr` which would result `12.4 kr kr` also I could send string like `123,44` (with comma instead of dot) which would result in `123,44 ,-` which is incorrect. So instead of handling all of this it's better to just drop it and let channels send in the proper value if they are sending a string value.

PR #548, opened 2020-02-10
Fix/add size to heading

bogsti:


component-lib/src/atoms/Heading/Heading.d.ts
```diff
@@ -21,6 +21,10 @@ export interface HeadingProps {
    * Content of this heading.
    */
   children?: React.ReactNode;
+
+  size?: string;
```
nmenglund: This should only be the valid sizes, i.e. `size?: 'mega' | 'xxl' | ...` etc. Maybe it's a good idea to convert the whole component to TS?

component-lib/src/atoms/Heading/Heading.d.ts
```diff
@@ -21,6 +21,10 @@ export interface HeadingProps {
...
+  align?: 'left' | 'center' | 'right' | 'justify';
```
nmenglund: Similar to the discussion on whether `color` was an appropriate attribute on the `Icon` component, I don't think the _align_ prop belongs on the Heading component. I think it would be more intuitive and consistent to use CSS classes or e.g. `<Heading style={{ textAlign: 'right' }} />` when the styleguide consumers want to alter the default styles

nmenglund: That said --- `className` _should_ be a prop, and while its read by the component JSX code, it's not defined in the TS type.

PR #551, opened 2020-02-12
feat/icon-link-with-new-icons

bogsti:
Screenshot 2020-02-12 at 16 28 53


component-lib/src/atoms/IconLink/IconLink.d.ts
```diff
@@ -7,6 +7,9 @@ export interface IconLinkProps {
    * One of 'download', 'linkedIn' or 'twitter'.
    */
   iconName?: IconLinkIconName;
+  icon?: string;
```
nmenglund: Should be `IconDefinition`, not string

nmenglund:
Would it make sense to convert the whole IconLink component to TS? This is preferable to avoid having to define the TS types in .d.ts separately from the component itself.

PR #569, opened 2020-02-25
Molecule: Hero. TypeScript + Storybook.

alexyork:
Not ready for merge. I am looking for feedback as to if this is done correctly, before I do more molecules.


component-lib/src/molecules/Hero/Hero.tsx
```diff
@@ -34,96 +33,98 @@ import uniqueId from 'lodash/uniqueId';
...
-        <picture className="hero__logo">
-          {logoSources.map(it => (
+
+interface HeroSources {
+  type?: 'mobile' | 'desktop';
+  srcSet?: string;
+}
+
+export interface HeroProps {
+  className: string;
```
nmenglund: Should probably be optional: `className?: string;`

component-lib/src/molecules/Hero/Hero.tsx
```diff
@@ -34,96 +33,98 @@ import uniqueId from 'lodash/uniqueId';
...
+  heading: string;
+  subheading: string;
+  alt: string;
+  logoSources?: any[];
```
nmenglund: Should also be `HeroSources[];`?

component-lib/src/molecules/Hero/Hero.tsx
```diff
@@ -34,96 +33,98 @@ import uniqueId from 'lodash/uniqueId';
...
+  return (
+    <a
+      className={classnames('hero heading-link', {
+        [className]: className,
+        [`hero--with-pebbles hero--with-pebbles-${pebbles}`]: pebbles,
+        [`hero--with-pebbles-only-on-${mode}`]: mode,
+      })}
+      {...rest}
+    >
+      <React.Fragment>
```
nmenglund: Pointless fragment; already contained by `<a>`
(I see this was in the old code too, but might as well clean up a bit ...)

component-lib/src/molecules/Hero/Hero.tsx
```diff
@@ -34,96 +33,98 @@ import uniqueId from 'lodash/uniqueId';
...
+              className={classnames('hero__logo-image', {
+                [`hero__logo-image--align-${logoAlign || 'top-right'}`]: logoAlign,
+              })}
+              alt={alt}
+            />
+          </picture>
+        )}
+        <div
+          className={classnames('hero__box heading-link--focus-area', {
+            [`hero__box--align-${alignBox || 'left'}`]: alignBox,
```
nmenglund: If `alignBox` is falsy, the object property will have a falsy value and no css class will be added (thus, `|| 'left'` is pointless)

If `left` should really be the default, I'd rather do `
```
className={classnames(
  'hero__box heading-link--focus-area',
  `hero__box--align-${alignBox || 'left'}`
)}
```

component-lib/src/molecules/Hero/Hero.tsx
alexyork: Agree + fixed.

component-lib/src/molecules/Hero/Hero.tsx
alexyork: Agree + fixed.

And yes, this was already there from earlier. My goal with this PR was to help with the migration to TypeScript & Storybook, not make any functional changes. Although I agree, small cleanups here and there are a good thing, so I can do them when I see them.

component-lib/src/molecules/Hero/Hero.tsx
alexyork: Agree + fixed. Again, it was like this in the old code, and I wasn't focusing on fixing, just migrating to TS + Storybook.

component-lib/src/molecules/Hero/Hero.tsx
alexyork: Agree + fixed + a similar fix in one other place in the file.

alexyork:
@nmenglund I fixed all the issues, and amended my commit and force-pushed.

@bogsti @nmenglund My thought with this PR was to support the migration to TypeScript (which I love, and think is a great idea for something like a styleguide) and Storybook (again, great idea). I basically just moved/migrated code - made no changes functionally. If you guys want me to migrate more components, in the same way I did this one, let me know and I will continue 👍

PR #570, opened 2020-02-25
Create draft subscription type in ShoppingCartItem

alfredbirk:


src/components/examples/molecules/ShoppingCart/Default.js
```diff
@@ -39,7 +39,7 @@ const cart = [
     subtitle: 'Telia SVITSJ i 24 md.',
     items: [
       {
-        type: 'SUBSCRIPTION',
+        type: 'SUBSCRIPTION_DRAFT',
```
joakikr: Should probably change the `subtitle` as well so the example is correct after designs.

PR #572, opened 2020-02-27
WIP feat/new-icons-in-links-n-lists

bogsti:


component-lib/src/atoms/Link/Link.tsx
```diff
@@ -0,0 +1,92 @@
...
+export type LinkIcon = 'forward-icon' | 'back-icon';
+
+export type LinkIconPosition = 'before' | 'after';
+
+export interface LinkProps {
+  className?: string;
+  /**
+   * Content of this link.
+   */
+  text?: React.ReactNode;
```
nmenglund: We need this for backward compatability, but maybe we should add a @deprecated here to nudge people over to using `children` instead?

component-lib/src/atoms/Link/Link.tsx
```diff
@@ -0,0 +1,92 @@
...
+  const {
+    className,
+    text,
+    children,
+    href,
+    icon,
+    iconPosition,
+    iconColor = 'core-purple',
+    negative,
+    target = '_self',
```
nmenglund: It might be clearer to just leave target as `undefined` if it is not provided?

component-lib/src/atoms/Link/Link.tsx
```diff
@@ -0,0 +1,92 @@
...
+   * Content of this link.
+   */
+  children?: React.ReactNode;
+
+  target?: string;
+
+  /**
+   * Choose between forward arrow (forward-icon) and backward arrow (back-icon) icon
+   */
+  icon?: IconDefinition;
```
nmenglund: If we want to lock this down, we can use `icon?: 'arrow-left' | 'arrow-right';` instead, that way it maps directly to IconDefinition. Unless we want users to be able to provide any icon ...

nmenglund: And the comment above should be updated

component-lib/src/atoms/Link/Link.tsx
```diff
@@ -0,0 +1,92 @@
...
+ * Status: *finished*.
+ * Category: Links
+ *
+ * The target attribute is set to '_self' which is the default value if the attribute is not specified.
+ * If you rather prefer to open the linked document in a new browser window or tab, you can set the target to '_blank' instead.
+ *
+ * Our links have _two visual means_ of being identified: an underline and bolder font.
+ */
+
+export type LinkIcon = 'forward-icon' | 'back-icon';
```
nmenglund: Not using this right now, might as well remove it.

component-lib/src/atoms/Link/Link.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+export type LinkIconPosition = 'before' | 'after';
```
nmenglund: Might as well move this into the prop definition

PR #582, opened 2020-03-13
Fix shoppingCartItem borders

alfredbirk:
Removed code which changes border radius on only parts of the ShoppingCartItem when changing screen size. Not quite sure why this code was there in the first place. Borders are still not perfect since the dashed individual borders overlap slightly.

### DIFF
--- a/component-lib/src/molecules/ShoppingCart/ShoppingCartTable.pcss
+++ b/component-lib/src/molecules/ShoppingCart/ShoppingCartTable.pcss
@@ -25,14 +25,6 @@
                 padding: 0.5rem 1rem 0.5rem 0;
                 width: initial;
             }
-
-            &:first-child {
-                border-radius: 0.3rem 0 0 0.3rem;
-            }
-
-            &:last-child {
-                border-radius: 0 0.3rem 0.3rem 0;
-            }
         }
 
         &__cell,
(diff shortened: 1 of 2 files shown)


component-lib/src/molecules/ShoppingCart/ShoppingCartTable.pcss
```diff
@@ -25,14 +25,6 @@
...
                 width: initial;
             }
-
-            &:first-child {
-                border-radius: 0.3rem 0 0 0.3rem;
-            }
-
-            &:last-child {
-                border-radius: 0 0.3rem 0.3rem 0;
-            }
```
joakikr: Without this CSS the item no longer has slightly rounded corners - which is a part of the design. Not sure if we can remove this.

alfredbirk: Okay, how about now? I added it back, but also added the same border-radius to all children of dashed items so it stays the same when zooming.

joakikr: Screen Shot 2020-03-13 at 14 21 48

If zooming in one can now see the rounded corners on the elements in the middle. This is a tricky one - if addinng rounded corners to the outside elements they get different dashing on some levels of zoom, or, if adding rounded corners on all elements the line is not completely straight 🤔

joakikr: Maybe just keep it as is since it's only a "problem" when zooming on page? Or maybe check with UX if we can remove the `border-radius`.

alfredbirk: Alright, probably better to keep it as it as

PR #590, opened 2020-03-19
Fix ShoppingCartTable sorting

joakikr:


joakikr:
> An alternative to assigning each type a number is to have them in an array and use the indices for sorting

That will work to. I think we can just keep it as an object for now👍

PR #596, opened 2020-04-07
Remove knobs in subscription stories

alfredbirk:


joakikr:
A full subscription listing page already exists under `?path=/docs/examples-pages--subscription-sample-page`. 

The intention here was to move all examples under `/src/components/examples/molecules/Subscription` to exports from the `Subscription.stories.tsx` file. See for instance `Alert.stories.tsx` for guidance.

alfredbirk:
> A full subscription listing page already exists under `?path=/docs/examples-pages--subscription-sample-page`.
> 
> The intention here was to move all examples under `/src/components/examples/molecules/Subscription` to exports from the `Subscription.stories.tsx` file. See for instance `Alert.stories.tsx` for guidance.

Okay, I have misunderstood. Moved the examples to storybook now!

PR #600, opened 2020-04-28
fix/small-updates

bogsti:
- Storybook organizing sidebar
- Link table border bottom removed
- Link padding removal on links without icon

### DIFF
--- a/component-lib/src/business/Checkbox/Checkbox.stories.tsx
+++ b/component-lib/src/business/Checkbox/Checkbox.stories.tsx
@@ -5,7 +5,7 @@ import { withKnobs, text, boolean } from "@storybook/addon-knobs";
 
 export default {
     component: Checkbox,
-    title: "Business/Checkbox",
+    title: "Business|Checkbox",
     decorators: [withKnobs]
 }
 
--- a/component-lib/src/business/Table/Table.stories.tsx
+++ b/component-lib/src/business/Table/Table.stories.tsx
@@ -7,7 +7,7 @@ import flow from 'lodash/fp/flow';
 
 export default {
     component: Table,
-    title: 'Business/Table'
+    title: 'Business|Table'
 }
 
 const subscribers = require('./sampledata.json');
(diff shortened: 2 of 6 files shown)


component-lib/src/business/Checkbox/Checkbox.stories.tsx
```diff
@@ -5,7 +5,7 @@ import { withKnobs, text, boolean } from "@storybook/addon-knobs";
 
 export default {
     component: Checkbox,
-    title: "Business/Checkbox",
+    title: "Business|Checkbox",
```
nmenglund, line 8: We should probably stick with `'/'` -- storybook is converging on using '/' as the single hierarchy separator. See https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#simplified-hierarchy-separators

component-lib/src/business/Table/Table.stories.tsx
```diff
@@ -7,7 +7,7 @@ import flow from 'lodash/fp/flow';
 
 export default {
     component: Table,
-    title: 'Business/Table'
+    title: 'Business|Table'
```
nmenglund, line 10: We should probably stick with `'/'` -- storybook is converging on using '/' as the single hierarchy separator. See https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#simplified-hierarchy-separators

component-lib/src/stories/home/ColorPalette.stories.tsx
```diff
@@ -3,7 +3,7 @@ import colorsJSON from '../lib/colorsJSON';
 import { colors, Icon } from '../../index';
 
 export default {
-  title: 'Introduction|Home/Color Palette',
+  title: 'Introduction|Color Palette',
```
nmenglund, line 6: We should probably stick with `'/'` -- storybook is converging on using '/' as the single hierarchy separator. See https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#simplified-hierarchy-separators

PR #603, opened 2020-05-13
Add hook and fix paging

antons1:
Adds a custom hook for table state that can be used, also fixes paging to no longer be dependent on a page number, but only from/to props. So that, if you see from 11-20 and switch to 25 per page, you no longer jump to see from 26-50, but instead see 11-36.

### DIFF
--- a/component-lib/src/business/Table/Table.tsx
+++ b/component-lib/src/business/Table/Table.tsx
@@ -240,25 +240,25 @@ type TablePagingControlsProps = {
   to: number;
   perPage: number;
   dataLength: number;
-  amountOfSelectedRows?: number;
-  selectedRowsString?: string
+  numberOfSelectedRows?: number;
+  selectedRowsLabel?: string
 
   onPerPageChange: (perPage: number, e?: React.ChangeEvent<HTMLSelectElement>) => void;
   onPageChange: (forward: boolean, e?: React.MouseEvent<HTMLButtonElement>) => void;
 
-  fromToString?: string;
-  perPageString?: string;
+  fromToLabel?: string;
+  perPageLabel?: string;
   selectOptions?: Array<number>;
 };
 
 export const TablePagingControls: React.FC<TablePagingControlsProps> = props => {
   return (
     <form onSubmit={e => e.preventDefault()} className="table-paging">
-        {props.amountOfSelectedRows ? <span className="table-paging__text">
-            {`${props.amountOfSelectedRows} ${props.selectedRowsString || "rader er valgt"}`}
+        {props.numberOfSelectedRows ? <span className="table-paging__text">
+            {`${props.numberOfSelectedRows} ${props.selectedRowsLabel || "rader er valgt"}`}
         </span> : null}
       <label className="table-paging__text">
-        {props.perPageString || 'Rader per side: '}
+        {props.perPageLabel || 'Rader per side: '}
         <select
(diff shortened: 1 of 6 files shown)


component-lib/src/business/Table/Table.tsx
```diff
@@ -240,11 +240,11 @@ type TablePagingControlsProps = {
   to: number;
   perPage: number;
   dataLength: number;
-  page: number;
-  maxPage: number;
+  amountOfSelectedRows?: number;
```
HeleneGrini, line 243: numberOfSelectedRows ?

component-lib/src/business/Table/Table.tsx
```diff
@@ -240,11 +240,11 @@ type TablePagingControlsProps = {
...
+  selectedRowsString?: string
```
HeleneGrini, line 244: kanskje kalle dette en tittel/label?

component-lib/src/business/Table/tableHooks.ts
```diff
@@ -0,0 +1,185 @@
...
+    type: 'SELECT_ROW' | 'DESELECT_ROW',
+    id: string | number
+} | {
+    type: 'SELECT_ALL' | 'DESELECT_ALL'
+}
+
+type TableStateHookType = [
+    TableState,
+    {
+        changeSorting: changeSortingAction,
```
HeleneGrini, line 49: Would maybe call these setXYZ, just by convention

component-lib/src/business/Table/tableHooks.ts
```diff
@@ -0,0 +1,185 @@
...
+const newFrom = (currentFrom: number, forward: boolean, perPage: number, dataLength: number) => {
+    const diff = forward ? perPage : perPage * -1;
+    return currentFrom + diff > dataLength ? dataLength : currentFrom + diff < 0 ? 0 : currentFrom + diff;
+}
+const maybePagedData = ({ data, paging, from, to }: TableState) => paging ? data.slice(from, to) : data;
+
+function tableReducer(state: TableState, action: tableStateAction): TableState {
+    switch (action.type) {
+        case "CHANGE_PAGE":
+            return _.cloneDeep({
```
HeleneGrini, line 70: Do you need to cloneDeep here, when you spread the state?

component-lib/src/business/Table/tableHooks.ts
```diff
@@ -0,0 +1,185 @@
...
+        | {
+            selecting: true,
+            selectionId: string | number,
+            allSelected?: boolean,
+            selectedRows?: Array<string | number>
+        } | {
+        }
+    )
+
+export function useTableState(args: TableStateArgs): TableStateHookType {
```
HeleneGrini, line 156: would also call the file the same as the hook

component-lib/src/business/Table/tableHooks.ts
HeleneGrini, line 49: would maybe call these setXYZ, just by convention

component-lib/src/business/Table/tableHooks.ts
antons1: Since spread only clones 1 level down, I usually default to cloning the object in cases where the state contains more deeply nested data, such as in this state where `data` is probably an object with more than one level. I am not sure if it is strictly necessary.

PR #619, opened 2020-07-01
Add business tabs to component lib

HeleneGrini:


component-lib/src/business/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,40 @@
+import * as React from 'react';
+import cs from 'classnames';
+import { TabProps } from './index';
```
nmenglund, line 3: Prefer to import directly from the file containing the definition when in the same dir (avoid unnecessary indirection)

component-lib/src/business/Tabs/index.tsx
```diff
@@ -0,0 +1,2 @@
+export * from './Tabs';
```
nmenglund, line 1: There's an element of preference here, but I think we should prefer explicit exports to ensure control over the exposed API

PR #624, opened 2020-07-14
Added ColorPicker component

joakikr:
Screen Shot 2020-07-14 at 15 42 06

### DIFF
--- a/component-lib/src/atoms/ColorPicker/ColorPicker.d.ts
+++ b/component-lib/src/atoms/ColorPicker/ColorPicker.d.ts
@@ -1,0 +1,17 @@
+import * as React from 'react';
+
+export interface Color {
+    id: string;
+    color: string;
+    name: string;
+}
+
+export interface ColorPickerProps {
+    selected?: Color | null;
+    colors: Color[];
+    onSelect?: (...args: any[]) => any;
+}
+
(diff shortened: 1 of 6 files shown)


component-lib/src/atoms/ColorPicker/ColorPicker.d.ts
```diff
@@ -0,0 +1,17 @@
...
+export interface Color {
+    id: string;
+    color: string;
+    name: string;
+}
+
+export interface ColorPickerProps {
+    selected?: Color | null;
+    colors: Color[];
+    onSelect?: (...args: any[]) => any;
```
ropimo, line 12: Could you narrow it down?
maybe: `(color: Color) => void`

component-lib/src/atoms/ColorPicker/ColorPicker.d.ts
```diff
@@ -0,0 +1,17 @@
+import * as React from 'react';
+
+export interface Color {
+    id: string;
+    color: string;
+    name: string;
+}
+
+export interface ColorPickerProps {
+    selected?: Color | null;
```
ropimo, line 10: that's more of a coding styling, but is it needed to support both `undefined` and `null` ?

PR #629, opened 2020-07-21
Added BigImageDialog component

joakikr:
Depends on https://github.com/TeliaSoneraNorge/styleguide/pull/624.

Added `ProductModalDialog` component for showing a recommended product in a modal.
Didn't want to alter `ModalDialog`, so borrowed much code (js and pcss) from it instead and created this component.

Examples for small and large screen:
Screen Shot 2020-07-21 at 13 19 54

Screen Shot 2020-07-21 at 13 20 07


component-lib/src/molecules/BigImageDialog/BigImageDialog.d.ts
```diff
@@ -0,0 +1,16 @@
+import * as React from 'react';
+
+export interface BigImageDialogProps {
+  name: string;
+  heading: string;
+  description?: string;
+  src: string;
+  children: React.ReactNode;
+  onClose?: (...args: any[]) => any;
```
ropimo, line 9: maybe `() => void` ?

component-lib/src/molecules/BigImageDialog/BigImageDialog.d.ts
```diff
@@ -0,0 +1,16 @@
...
+  standalone?: boolean;
+  renderTo?: any;
```
ropimo, line 11: should this be a `React.ReactNode` or something like that?

PR #631, opened 2020-07-29
Add sliding shopping cart

alfredbirk:


component-lib/src/organisms/SlidingShoppingCart/SlidingShoppingCart.jsx
```diff
@@ -0,0 +1,79 @@
...
+  hasPaid,
+  disclaimers,
+}) => {
+  const [shouldShowCart, setShouldShowCart] = useState(false);
+
+  return (
+    <>
+      <button onClick={() => setShouldShowCart(!shouldShowCart)} className="sliding-shopping-cart__cart-button">
+        <Icon icon="shoppingcart" className="sliding-shopping-cart__cart-icon" />
+      </button>
```
joakikr, line 34: This should probably not be a part of the slider. Better to take in a prop that indicate whether the cart should be shown or not.

component-lib/src/organisms/SlidingShoppingCart/SlidingShoppingCart.pcss
```diff
@@ -0,0 +1,52 @@
...
+        background-color: inherit;
+        cursor: pointer;
+        &:focus {
+            outline: 0;
+        }
+    }
+    &__cart-icon {
+        color: var(--core-purple);
+    }
+    &__container {
```
joakikr, line 13: This container should scale better with the cart. Can have `50vw`, but should probably also have a `min-width` that still show the whole row of a cart item.

Also, if cart is longer than page it's not possible to scroll.

component-lib/src/organisms/SlidingShoppingCart/SlidingShoppingCart.pcss
```diff
@@ -0,0 +1,52 @@
...
+        }
+    }
+    &__checkout-button-container {
+        display: flex;
+        justify-content: flex-end;
+    }
+    &__checkout-button {
+        margin: 1rem 1rem 0 0 !important;
+    }
+    &__close-button {
```
joakikr, line 37: Add pointer cursor. Also need to be visible when element is active/in focus somehow. Can maybe add the scaling for these pseudos as well?

component-lib/src/organisms/SlidingShoppingCart/index.d.ts
```diff
@@ -0,0 +1 @@
+export { default } from './SlidingShoppingCart';
```
joakikr, line 1: Instead of having `index.d.ts`, `SlidingShoppingCart.ts` and `SlidingShoppingCart.jsx` you can just have `SlidingShoppingCart.tsx` I think.

You can then define the prop interface in top of that file, and it should probably be more descriptive than just the `any` type. Have a look at `Button.tsx` for an example.

alfredbirk: Does this mean I need to define all those interfaces you defined for cart also in my file as one of the props is `cartItems`?

joakikr: Yes! Can let that be an own task 👍

joakikr: So just create a prop interface for all the props and let `cartItems` be of type `any` for now.

PR #641, opened 2020-08-11
Add optional cart icon to menu

alfredbirk:


component-lib/src/molecules/Menu/MenuCart.pcss
```diff
@@ -0,0 +1,64 @@
...
+    &__icon {
+        display: block;
+
+        height: var(--cart-icon-small);
+        width: var(--cart-icon-small);
+
+        @media screen and (min-width: 42em) {
+            height: var(--cart-icon-large);
+            width: var(--cart-icon-large);
+        }
```
joakikr, line 33: Size scaling should probably match the search icon breakpoints and scaling

component-lib/src/molecules/Menu/MenuCart.pcss
```diff
@@ -0,0 +1,64 @@
...
+        letter-spacing: 1px;
+        padding: 0.7rem 0.7rem;
+        position: relative;
+
+        &:hover, &:focus {
+            outline: none;
+            > .menu-cart__icon * {
+                fill: var(--core-purple);
+            }
+        }
```
joakikr, line 22: `cursor: pointer` on these states

component-lib/src/molecules/Menu/MenuCart.pcss
```diff
@@ -0,0 +1,64 @@
...
+    --cart-icon-small: 1.5rem;
+    --cart-icon-large: 2rem;
+    --cart-item-count-size: 1.3rem;
+    --cart-item-count-size-small: 1rem;
+}
+
+.menu-cart {
+    &__button {
+        display: flex;
+        background-color: transparent;
```
joakikr, line 11: Color should be core purple.

component-lib/src/molecules/Menu/MenuCart.pcss
```diff
@@ -0,0 +1,64 @@
...
+        width: var(--cart-icon-small);
+
+        @media screen and (min-width: 42em) {
+            height: var(--cart-icon-large);
+            width: var(--cart-icon-large);
+        }
+    }
+    &__item-count {
+        background-color: var(--red);
+        border-radius: calc(var(--cart-item-count-size-small) / 2);
```
joakikr, line 37: Just have `border-radius: 50%` always to create circle effect?

component-lib/src/organisms/SlidingShoppingCart/SlidingShoppingCart.pcss
```diff
@@ -22,7 +22,9 @@
         opacity:0;
         transition: 0.2s;
         overflow-y: auto;
+        z-index: -50;
```
joakikr, line 25: Why is this necessary?

alfredbirk: Since the opacity just is set to 0, the cart is actually still there, so it was not possible to click on things under it. Using `display: none` didn't work, cause it ruined the sliding animation. But it worked using `visibility: hidden` so I changed to that.

component-lib/src/molecules/Menu/MenuCart.pcss
alfredbirk: Talked to Jørgen about this - black is better cause it matches the other icons and then it can become purple on hover.

component-lib/src/organisms/SlidingShoppingCart/SlidingShoppingCart.pcss
joakikr: Ah ofc! Visibility attr is good then 👍

PR #660, opened 2020-08-26
Recursively get all dropdown items to support use of wrapper components

HeleneGrini:


component-lib/src/molecules/Dropdown/Dropdown.stories.tsx
```diff
@@ -65,7 +65,7 @@ const countries = [
   { name: 'Zimbabwe' },
 ];
 
-storiesOf('Business/Dropdown', module)
+storiesOf('Component library|Molecules/Dropdown', module)
```
nmenglund, line 68: Always use `/` as separator in Storybook 6.0 -- i.e. `Component library/Molecules/Dropdown`

component-lib/src/molecules/Dropdown/DropdownItem.tsx
```diff
@@ -55,6 +53,12 @@ export const DropdownItem: React.FC<DropdownItemProps> = props => {
   const { open, toggle, highlightIndex } = useDropdownContext();
   const itemRef = useRef<HTMLButtonElement>(null);
 
+  useEffect(() => {
+    if (props.index === highlightIndex && itemRef.current) {
+      itemRef.current.scrollIntoView({ block: 'nearest', inline: 'nearest' });
```
nmenglund, line 58: Tested this in IE11? It seems IE11 only supports the boolean parameter type: https://caniuse.com/#search=scrollIntoView

nmenglund: (do we still need to support IE11? maybe put all IE11 polyfills a bucket script somewhere?)

nmenglund (changes requested):

component-lib/src/molecules/Dropdown/useAccessibleDropdown.ts
```diff
@@ -42,15 +42,17 @@ export const useAccessibleDropdown = () => {
       }
     };
     const navigate = (e: KeyboardEvent) => {
+      const arrowDown = 40;
+      const arrowUp = 38;
       if (ctx.open && e.target && ctx.menuRef.current) {
-        if (e.key === 'ArrowDown') {
+        if (e.keyCode === arrowDown) {
```
nmenglund, line 48: I think the `const` definition is unnecessary; a `// arrow down` comment would suffice

component-lib/src/molecules/Dropdown/DropdownItem.tsx
```diff
@@ -53,6 +53,13 @@ export const DropdownItem: React.FC<DropdownItemProps> = props => {
   const { open, toggle, highlightIndex } = useDropdownContext();
   const itemRef = useRef<HTMLButtonElement>(null);
 
+  /**
+   * Work in progress
+   *
+   * Currently works for dropdown menus without search item(s).
+   * NB: Not supported properly in ie11 from the styleguiden. This currently
+   * needs to be handled by the user
```
nmenglund, line 61: I think you should elaborate a little bit in the comment; _what_ isn't supported? _How_ do users need to handle it?

("NB: `scrollIntoView` is not supported properly in IE11`", "needs to be handled by the user with a polyfill")

nmenglund (changes requested):

nmenglund (approved):

antons1 (approved):

PR #668, opened 2020-09-01
Add buttonIsProcessing prop to subscription accordion

alfredbirk:


component-lib/src/molecules/SubscriptionAccordion/SubscriptionAccordion.jsx
```diff
@@ -114,6 +116,7 @@ SubscriptionAccordion.propTypes = {
   onOpen: PropTypes.func,
   onSelect: PropTypes.func,
   buttonText: PropTypes.string,
+  buttonIsProcessing: PropTypes.bool,
```
joakikr, line 119: Add missing props to ts type definition 👍

PR #671, opened 2020-09-02
SubscriptionAccordion changes

joakikr:
Remove logic to render button.
This should be handled by consumer instead through children prop.

### DIFF
--- a/component-lib/src/molecules/SubscriptionAccordion/SubscriptionAccordion.jsx
+++ b/component-lib/src/molecules/SubscriptionAccordion/SubscriptionAccordion.jsx
@@ -1,4 +1,5 @@
 import React, { useRef, useEffect } from 'react';
+import PropTypes from 'prop-types';
 import cn from 'classnames';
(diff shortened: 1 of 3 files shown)


component-lib/src/molecules/SubscriptionAccordion/SubscriptionAccordion.jsx
```diff
@@ -77,51 +72,13 @@ const SubscriptionAccordion = ({
...
-  disclaimers: PropTypes.element,
-  scrollToOnOpen: PropTypes.bool,
-  className: PropTypes.string,
-  onOpen: PropTypes.func,
-  onSelect: PropTypes.func,
-  buttonText: PropTypes.string,
-  buttonIsProcessing: PropTypes.bool,
-  description: PropTypes.element,
-};
-
```
alfredbirk, line 124: Some clients don't use TS (like rapidshop), so maybe a good idea to keep proptypes?

alfredbirk (changes requested):

PR #684, opened 2020-09-10
Improve dropdown docs

HeleneGrini:
First and last commit are minor fixes. 
The rest is refactoring for readability, comments and story descriptions.


nmenglund (approved):
Yes, bedre :-) LGTM!

PR #693, opened 2020-09-22
declare className prop in all FullWidthTable components

ropimo:
The className prop was not declared in the FullWidthTable components. This PR should fix it.
This changes might affect all sites. I hope it doesn't break anything. Please verify it works.

I also added some Storybook knobs for it (and the white footer, to reduce code)

### DIFF
--- a/src/atoms/FullWidthTable/FullWidthTable.d.ts
+++ b/src/atoms/FullWidthTable/FullWidthTable.d.ts
@@ -1,29 +1,27 @@
 import * as React from 'react';
 import { UseThSortable } from './useThSortable';
 
-declare type CommonProps = React.HTMLAttributes<HTMLElement>;
-
-declare type FullWidthTableProps = CommonProps & {
+declare type FullWidthTableProps = React.HTMLAttributes<HTMLTableElement> & {
   clickable?: boolean;
 };
 
-declare type ThSortableProps = CommonProps & {
+declare type ThSortableProps = React.HTMLAttributes<HTMLTableHeaderCellElement> & {
   onSort: () => void;
   isSorted?: boolean;
   isSortedDesc?: boolean;
   sortPriority?: number;
 };
 
-declare type TFootProps = CommonProps & {
+declare type TFootProps = React.HTMLAttributes<HTMLTableSectionElement> & {
   white?: boolean;
 };
 
 declare const FullWidthTable: React.FC<FullWidthTableProps> & {
-  THead: React.FC<CommonProps>;
-  TBody: React.FC<CommonProps>;
-  Th: React.FC<CommonProps>;
-  Tr: React.FC<CommonProps>;
-  Td: React.FC<CommonProps>;
+  THead: React.FC<React.HTMLAttributes<HTMLTableSectionElement>>;
+  TBody: React.FC<React.HTMLAttributes<HTMLTableSectionElement>>;
+  Th: React.FC<React.HTMLAttributes<HTMLTableHeaderCellElement>>;
+  Tr: React.FC<React.HTMLAttributes<HTMLTableRowElement>>;
+  Td: React.FC<React.HTMLAttributes<HTMLTableDataCellElement>>;
 
   ThSortable: React.FC<ThSortableProps>;
 
--- a/src/atoms/FullWidthTable/FullWidthTable.d.ts
+++ b/src/atoms/FullWidthTable/FullWidthTable.d.ts
@@ -1,34 +1,40 @@
 import * as React from 'react';
 import { UseThSortable } from './useThSortable';
+import { MouseEventHandler } from 'react';
 
-declare type FullWidthTableProps = {
+declare type CommonProps = {
+  className?: string;
+};
+
+declare type FullWidthTableProps = CommonProps & {
   clickable?: boolean;
 };
 
-declare type TrProps = {
-  onClick: () => void;
+declare type TrProps = CommonProps & {
+  onClick?: MouseEventHandler;
 };
 
-declare type ThSortableProps = {
+declare type ThSortableProps = CommonProps & {
+  onSort: () => void;
   isSorted?: boolean;
   isSortedDesc?: boolean;
   sortPriority?: number;
 };
 
-declare type TFootProps = {
+declare type TFootProps = CommonProps & {
   white?: boolean;
 };
 
-declare const FullWidthTable: React.FC<React.ReactNode | FullWidthTableProps> & {
-  THead: React.FC<React.ReactNode>;
-  TBody: React.FC<React.ReactNode>;
-  Th: React.FC<React.ReactNode>;
-  Tr: React.FC<React.ReactNode | TrProps>;
-  Td: React.FC<React.ReactNode>;
+declare const FullWidthTable: React.FC<FullWidthTableProps> & {
+  THead: React.FC<CommonProps>;
+  TBody: React.FC<CommonProps>;
+  Th: React.FC<CommonProps>;
+  Tr: React.FC<TrProps>;
+  Td: React.FC<CommonProps>;
 
-  ThSortable: React.FC<React.ReactNode | ThSortableProps>;
+  ThSortable: React.FC<ThSortableProps>;
 
-  TFoot: React.FC<React.ReactNode | TFootProps>;
+  TFoot: React.FC<TFootProps>;
 
   useThSortable: UseThSortable;
 };
--- a/src/atoms/FullWidthTable/FullWidthTable.d.ts
+++ b/src/atoms/FullWidthTable/FullWidthTable.d.ts
@@ -1,17 +1,10 @@
 import * as React from 'react';
 import { UseThSortable } from './useThSortable';
-import { MouseEventHandler } from 'react';
 
-declare type CommonProps = {
-  className?: string;
-};
+declare type CommonProps = React.HTMLAttributes<HTMLElement>;
 
 declare type FullWidthTableProps = CommonProps & {
   clickable?: boolean;
-};
-
-declare type TrProps = CommonProps & {
-  onClick?: MouseEventHandler;
 };
 
 declare type ThSortableProps = CommonProps & {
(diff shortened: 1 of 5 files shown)


ropimo:
Screenshot 2020-09-22 at 16 14 40

src/atoms/FullWidthTable/FullWidthTable.stories.pcss
```diff
@@ -0,0 +1,11 @@
+.purple {
```
nmenglund, line 1: All .pcss files will be merged into the same, standard distribution bundle in the same shared CSS namespace (also in the npm package).

For stories, I suggest we avoid .pcss files altogether, and use inline styles where we need them (alternatively a shared, storybook-specific .pcss that's explicitly excluded from the main distribution)

src/atoms/FullWidthTable/FullWidthTable.stories.tsx
```diff
@@ -9,35 +9,45 @@ export default {
 
 export const Default = () => {
   const { THead, TBody, Th, Tr, Td } = FullWidthTable;
+  const column2classNames = optionsKnob(
```
nmenglund, line 12: Might replace this with something like `const styles = { purple: { color: colors.corePurple }, italic: { fontStyle: 'italic'; },` etc.
and use the option as lookup key?

src/atoms/FullWidthTable/FullWidthTable.stories.tsx
```diff
@@ -9,35 +9,45 @@ export default {
...
+    .replaceAll(',', ' ');
 
   return (
-    <FullWidthTable>
+    <FullWidthTable className={'no-ts-error'}>
       <THead>
         <Tr>
           <Th>Column 1</Th>
-          <Th>Column 2</Th>
+          <Th className={column2classNames}>Column 2</Th>
```
nmenglund, line 28: `<Th styles={selectedStyle}>` instead?

src/atoms/FullWidthTable/FullWidthTable.d.ts
```diff
@@ -1,34 +1,40 @@
 import * as React from 'react';
 import { UseThSortable } from './useThSortable';
+import { MouseEventHandler } from 'react';
 
-declare type FullWidthTableProps = {
+declare type CommonProps = {
+  className?: string;
```
nmenglund, line 6: `styles={}` might be useful here too?

src/atoms/FullWidthTable/FullWidthTable.d.ts
```diff
@@ -1,34 +1,40 @@
...
-declare const FullWidthTable: React.FC<React.ReactNode | FullWidthTableProps> & {
-  THead: React.FC<React.ReactNode>;
-  TBody: React.FC<React.ReactNode>;
-  Th: React.FC<React.ReactNode>;
-  Tr: React.FC<React.ReactNode | TrProps>;
-  Td: React.FC<React.ReactNode>;
+declare const FullWidthTable: React.FC<FullWidthTableProps> & {
+  THead: React.FC<CommonProps>;
+  TBody: React.FC<CommonProps>;
+  Th: React.FC<CommonProps>;
```
nmenglund, line 31: Will using only CommonProps cause a TS error if we use plain <th> attributes on `<Th>` (do we _want_ to be able to use any other generic attributes on `<Th>`?

nmenglund (changes requested):

src/atoms/FullWidthTable/FullWidthTable.stories.pcss
ropimo: Good catch. I will see what can I do to avoid leaking the storybook only styles to the main distribution

src/atoms/FullWidthTable/FullWidthTable.stories.tsx
ropimo: The point of this PR is to accept className styles defined as CSS classes, instead of inline styles. That's why I am using the **className** (which is **class** in pure HTML)

src/atoms/FullWidthTable/FullWidthTable.d.ts
ropimo: yes, sure. And `id?: string;` let me see if I find a more generic React HTML attributes type

src/atoms/FullWidthTable/FullWidthTable.d.ts
ropimo: Before they were defined as: `Th: React.FC<React.ReactNode>;` I wasn't sure of the value of doing so. Have we **lost** any accepted attributes with these PR changes?

src/atoms/FullWidthTable/FullWidthTable.stories.tsx
ropimo: `optionsKnob` is the StoryBook knob which allows multiple selections. Nor sure I understand what you mean.

src/atoms/FullWidthTable/FullWidthTable.d.ts
ropimo: I'm inclined for this:
`declare type CommonProps = React.HTMLAttributes<HTMLElement>;`

src/atoms/FullWidthTable/FullWidthTable.d.ts
ropimo: I think this will be the most generic way to solve it:
`declare type CommonProps = React.HTMLAttributes<HTMLElement>;`

src/atoms/FullWidthTable/FullWidthTable.stories.pcss
ropimo: By using .css instead of .pcss the styles will NOT end up in the generated Telia Styleguide final CSS

src/atoms/FullWidthTable/FullWidthTable.stories.tsx
nmenglund: Ah, I made a suggestion that wasn't particular to this PR -- my suggestion was to avoid className altogether so we could avoid an external .css file. This is fine!

src/atoms/FullWidthTable/FullWidthTable.stories.tsx
```diff
@@ -9,35 +9,45 @@ export default {
...
+    {
+      display: 'check',
+    }
+  )
+    .toString()
+    .replaceAll(',', ' ');
 
   return (
-    <FullWidthTable>
+    <FullWidthTable id={'defaultFullWidthTable'} className={'no-ts-error'}>
```
nmenglund, line 24: Prefer quoted strings over `{}` expressions when providing static strings as props ( `id="defaultFullWidthTable" className="no-ts-error"`)

src/atoms/FullWidthTable/FullWidthTable.stories.tsx
nmenglund: Related to the other comment, I meant continuing to use `optionsKnobs`, but provide a map of inline styles instead of classes. This (like the other comment) was a suggestion to avoid using a separate .css file. But this is fine now :+1:

src/atoms/FullWidthTable/FullWidthTable.d.ts
```diff
@@ -1,34 +1,33 @@
 import * as React from 'react';
 import { UseThSortable } from './useThSortable';
 
-declare type FullWidthTableProps = {
-  clickable?: boolean;
-};
+declare type CommonProps = React.HTMLAttributes<HTMLElement>;
```
nmenglund, line 4: Instead of a shared `CommonProps`, we can add the element-specific attributes to each element, e.g.:

```
declare type ThSortableProps = React.ThHTMLAttributes<HTMLTableHeaderCellElement> & { isSorted?: boolean; };
```

etc.

src/atoms/FullWidthTable/FullWidthTable.d.ts
```diff
@@ -1,34 +1,33 @@
...
 
-declare type TrProps = {
-  onClick: () => void;
+declare type FullWidthTableProps = CommonProps & {
+  clickable?: boolean;
 };
 
-declare type ThSortableProps = {
+declare type ThSortableProps = CommonProps & {
+  onSort: () => void;
```
nmenglund, line 11: Should `onSort` be optional here? Even in cases where it doesn't make sense to leave out a event handler, I often prefer making it optional to avoid having to fill it out with a temporary value during development

nmenglund (changes requested):

src/atoms/FullWidthTable/FullWidthTable.d.ts
ropimo: Making it optional would make it semantically incorrect, and therefore misleading. I strongly suggest you not making the props optional if they are mandatory. You can always add a `// @ts-ignore` on your client code if you don't want to use the advantages of a typed system.

src/atoms/FullWidthTable/FullWidthTable.stories.tsx
ropimo: I do prefer consistency myself and always using `attribute={***}` in the React code. But I will adapt it to the project standard.

src/atoms/FullWidthTable/FullWidthTable.d.ts
ropimo: I considered it myself, but I thought it was giving very little value, and was enough to make the previous code correct. But you are right, it's even more precise this way, so I will change it.

src/atoms/FullWidthTable/FullWidthTable.d.ts
nmenglund: Oh, come on now -- this is _not_ about not using the advantages of a typed system.

More along the lines of a `Button` -- its primary function is to be clicked, and it is rather pointless without an `onClick` (not taking HTML/forms into account here). Still, allowing for `onClick` to be not present or undefined reduces friction during development; for example if I want to place the button and test the layout before implementing the click handler. Or if I want to propagate a function that might be undefined from the "outside" into the event handler.

I often shamelessly prefer being pragmatic over being entirely semantically correct. I much prefer `{ isSorted?: boolean; isSortedDesc?:boolean; sortPriority?: number; }` over elaborate and cumbersome types that correctly define all valid combinations of props (e.g. prevent `isSortedDesc`,`sortPriority` from being provided when `isSorted` is false).

While this is an interesting point of discussion, I'm happy about the PR either way :)

nmenglund (approved):

PR #696, opened 2020-09-28
Add support for only applying focus style when tabbing 

HeleneGrini:


HeleneGrini:
With this solution we will have a new event listener for each component. As an alternative solution, we could make a component   whos only job is to detect tabbing and set a classname on the body (or we could expose the tab detection function). And perhaps remove base styles for focus, and sets our custom focus style when tabbing -- or the components could have responsibility for its own focus style. With this, each app will hava to render the component (or call the function)  at the root of the application. What do you think?

nmenglund:
I think a new event listener for each component is excessive! I think we _can_ consider the styleguide more than a simple collection of components, but I still think everything used from the styleguide should be explicitly imported from it. So I would suggest either exposing a global tab detection function (enableTabKeyDetector, like we have in Min Portal -- which I think is cleanest), or as a React hook (useTabKeyDetection) on the top-most component or a React component on the topmost component (`<App> <TabKeyDetection /> .... </App>`). Regardless of choice, it should be documented well, with examples

src/utils/useTabKeyDetection.ts
```diff
@@ -0,0 +1,48 @@
...
+ * (using the keyboard to navigate).
+ *
+ * We enable this by setting a classname on the body-element
+ * that removes focus style.
+ *
+ * When the user starts tabbing, we remove the class in order
+ * to apply focus style again.
+ *
+ * For yout application to supprt this,
+ * call the hook og render the component, at the root of
```
antons1, line 15: `og` => `or` eller `and`?

HeleneGrini: skla være `or` 😄  fikser!

src/utils/useTabKeyDetection.ts
```diff
@@ -0,0 +1,48 @@
...
+export function useTabKeyDetection() {
+  useEffect(() => {
+    function mouseDownHandler() {
+      document.body.classList.add('telia-hide-focus');
+      window.removeEventListener('mousedown', mouseDownHandler);
+      window.addEventListener('keydown', tabKeyHandler);
+    }
+
+    function tabKeyHandler(e: KeyboardEvent) {
+      if (e.keyCode === 9) {
```
antons1, line 27: `KeyboardEvent.keyCode` er deprecated https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode
Ser ut til at enten `KeyboardEvent.code` eller `KeyboardEvent.key` bør brukes. Har ikke satt meg inn i kompatibilitet for de to, så vet ikke om det er mulig å bruke.

src/utils/useTabKeyDetection.ts
```diff
@@ -0,0 +1,48 @@
...
+ * we only want to display focus styles when the user is tabbing
+ * (using the keyboard to navigate).
+ *
+ * We enable this by setting a classname on the body-element
+ * that removes focus style.
+ *
+ * When the user starts tabbing, we remove the class in order
+ * to apply focus style again.
+ *
+ * For yout application to supprt this,
```
mariannemagnussen, line 14: your, support

src/utils/useTabKeyDetection.ts
```diff
@@ -0,0 +1,48 @@
...
+ * call the hook or render the component, at the root of
+ * you application.
```
mariannemagnussen, line 16: your

mariannemagnussen (approved):
Ser bra ut! Noen typos bare :D 
Og som nevnt kunne vi kanskje holdt oss til et valg. så slipper man å ta stilling til det 🤷🏻‍♀️

mariannemagnussen (approved):

PR #722, opened 2020-10-23
Change hover effect for SubscriptionLinesAccordion

alfredbirk:


src/molecules/SubscriptionLinesAccordion/SubscriptionLinesAccordion.pcss
```diff
@@ -23,49 +23,37 @@
         width: 100%;
         padding: 1rem .75rem 0.2rem .75rem;
         color: black;
-        border: none;
+        border: 3px solid transparent;
+        border-left: 0px solid transparent;
+        border-right: 0px solid transparent;
```
joakikr, line 28: Should these be `3px` too to avoid the accordion from "jumping" on hover/click?

joakikr: I fixed it 👍

joakikr (approved):

PR #738, opened 2020-11-03
remove pcss import in chip component

HeleneGrini:


alexyork:
I haven't worked in the styleguide for a long time, so I don't have overview of how it works these days, or if this fix fixes the problem or not. I think @HeleneGrini and @mariannemagnussen are best placed to review, merge and ship this :-)

mariannemagnussen (approved):

PR #772, opened 2020-11-16
Fix clickable zone for SubscriptionLinesAccordion

alfredbirk:


src/molecules/SubscriptionLinesAccordion/SubscriptionLinesAccordion.jsx
```diff
@@ -115,12 +116,19 @@ const SubscriptionLinesAccordion = ({
...
         })}
-        onClick={e => handleOnOpen(e)}
       >
+        <button className="subscription-lines-accordion__choose-button" onClick={() => handleOnSelect()} />
+        <button
+          onClick={(e) => handleOnOpen(e)}
+          onMouseEnter={() => setIsHoveringExpand(true)}
+          onMouseLeave={() => setIsHoveringExpand(false)}
+          className="subscription-lines-accordion__expand-button"
+        />
```
joakikr, line 130: This is kind of hacky 😅  It also makes the tab order unatural it looks like. How about placing the content inside the `RadioButton` somehow, and then have the arrow icon outside?

Have a look at `DeliveryMethods` in RS where we do something similar I think.

alfredbirk: Might be true yes. Okay, I will try that

alfredbirk: Made a new PR for this: https://github.com/TeliaSoneraNorge/styleguide/pull/778

PR #843, opened 2021-01-05
Change onChange type in TextBoxWithLabel to play nice with event.target.value

kodeFant:
React.SyntheticEvent does not expose the necessary types for the input change event.


HeleneGrini:
I would suggest using the new TextField component, as is is from the new design system. I have deprecated the old TextBox components. Not sure how much work that would be for you. Is it manageable?  It would bring us closer to a common design system and cleaner Styleguide 💯 

Docs for the new TextField: https://styleguide.channelapi.telia.no/?path=/story/component-library-molecules-textfield--default

kodeFant:
Thanks, I think that will be manageable, will try that 😊

PR #861, opened 2021-01-20
Feature/welcome offer discount

agnethesoraa:


src/molecules/SubscriptionAccordion/SubscriptionAccordion.d.ts
```diff
@@ -23,6 +23,7 @@ export interface SubscriptionAccordionProps {
   price?: SubscriptionPrice;
   priceInfo?: string[];
   discount?: SubscriptionDiscount;
+  welcomeOfferDiscount?: SubscriptionDiscount;
```
joakikr, line 26: Instead of adding a new prop for this, maybe just add a color prop to `SubscriptionDiscount`?
It can have two values allowed (blue, green), and then css classes for each.

src/molecules/SubscriptionAccordion/SubscriptionAccordion.jsx
```diff
@@ -59,6 +60,14 @@ const SubscriptionAccordion = ({
         ) : (
           description
         )}
+        {welcomeOfferDiscount ? (
+          <div className="subscription-accordion__welcomeOfferDiscount">
+            <span className="subscription-accordion__welcomeOfferDiscount-price">-{welcomeOfferDiscount.price}</span>
```
joakikr, line 65: The `-` in the price can come from the prop itself 👍

src/molecules/SubscriptionAccordion/SubscriptionAccordion.d.ts
agnethesoraa: What do you suggest to do with the placement given that we only use a color prop?

src/molecules/SubscriptionAccordion/SubscriptionAccordion.d.ts
```diff
@@ -12,6 +12,7 @@ export type SubscriptionFeature = {
 export type SubscriptionDiscount = {
   price: SubscriptionPrice;
   description: string;
+  color?: string;
```
joakikr, line 15: Explicit say `color?: 'blue' | 'green'` to avoid passing in a color that doesnt exist?

joakikr (approved):

PR #868, opened 2021-01-22
Add discount row in shopping cart

agnethesoraa:


src/molecules/ShoppingCart/ShoppingCartTableFooter.jsx
```diff
@@ -70,6 +84,10 @@ ShoppingCartTableFooter.propTypes = {
   totalPriceUpfront: PropTypes.number,
   hasPaid: PropTypes.bool,
   formatPrice: PropTypes.func,
+  discount: PropTypes.shape({
+    label: PropTypes.string,
+    percentage: PropTypes.number,
```
joakikr, line 89: I think it's better not to pass in the percentage of discount from the consumer, but rather pass in the full discount price. In that way it's more versatile, say if the discount was not to be for the whole monthly payment 🙂

joakikr (commented):
Component looks good 👏  Just one comment to one of the props passed in.

src/molecules/ShoppingCart/ShoppingCartTableFooter.jsx
agnethesoraa: Aha, smart! I will change it to the full discount price! 👍

joakikr (approved):

PR #878, opened 2021-01-29
support additional content in step flow

HeleneGrini:
Any ideas on naming for that prop? 😄 

I also refactored the stories a bit and updated the background color according to the sketches.

### DIFF
--- a/src/business/StepFlow/StepFlow.pcss
+++ b/src/business/StepFlow/StepFlow.pcss
@@ -1,5 +1,5 @@
 .telia-step-flow {
-  background-color: white;
+  background-color: var(--grey-100);
   position: absolute;
   display: block;
   top: 0;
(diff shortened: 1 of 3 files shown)


mariannemagnussen (commented):
Nice! Man skal jo kunne sende inn hva som helst, så jeg har ikke noe bedre navn enn additionalComponent, tror det funker bra 😄   Siden den i storyen heter "withCart", kunne vi lagt på litt mer i prop beskrivelsen:  "Was originally added to support a shop cart" eller noe sånt?

hannaharnett:
Sweet! As I cannot come up with a better name, I agree with Marianne to just add a description 😄

hannaharnett (approved):

mariannemagnussen (approved):
👍

PR #883, opened 2021-02-08
Fix  height of business buttons

HeleneGrini:
I noticed that the buttons got different heights depending on how/where it was used, so i set the height to fixed according to the sketches


mariannemagnussen (approved):
👍

PR #888, opened 2021-02-10
add badge component

HeleneGrini:
Not sure if we should support positioning the badge relative to its child (eg. in the top right corner of the child, like a notification). I've looked into how other libs do it, and it seems to be a common practice. I can imagine it being useful for notifications etc. What do you think, should we support it? Might as well? 

Ref: https://material-ui.com/components/badges/#dot-badge

### DIFF
--- a/src/atoms/Badge/Badge.stories.tsx
+++ b/src/atoms/Badge/Badge.stories.tsx
@@ -1,0 +1,111 @@
+import React from 'react';
+import { Badge } from './Badge';
+import { colors } from '../../utils/colors';
+
+export default {
+  title: 'Component library/Atoms/Badge',
+  component: Badge,
+};
+
+export const Default = () => {
+  return (
+    <div>
+      <h3>States</h3>
+      <div style={{ display: 'flex' }}>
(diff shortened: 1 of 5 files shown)


mariannemagnussen:
I agree that we might as well implement this feature right away 👌

src/atoms/Badge/Badge.stories.tsx
```diff
@@ -0,0 +1,223 @@
...
+        </Badge>
+        <Badge kind="active" size="compact" position="bottom-right">
+          <Button label="Button" kind="secondary" size="compact" />
+        </Badge>
+        <Badge kind="active" size="compact" position="bottom-left">
+          <Button label="Button" kind="secondary" size="compact" />
+        </Badge>
+      </div>
+
+      <h4>Custom transform</h4>
```
mariannemagnussen, line 199:
```suggestion
      <h4>Custom position</h4>
```
?

mariannemagnussen (changes requested):
Nice! 👌  Hadde bare et forslag på ord

mariannemagnussen (approved):

PR #901, opened 2021-02-22
Update List with new style changes

mariannemagnussen:
- Remove shadow styling and replace it with outline (consistent with the Card component)
- Set inner div in ListItem to button if it is clickable
- Update border radius to 12px (will also be done for the Card component)
- Move expandedChildren to its own prop to keep using children as custom list items
- Remove hover effects when onClick is undefined

### DIFF
--- a/src/business/List/ListItem.tsx
+++ b/src/business/List/ListItem.tsx
@@ -38,14 +38,14 @@ export const ListItem: React.FC<ListItemProps & ListStyle> = (props) => {
         'telia-listItem--dark': color === 'dark',
         'telia-listItem--medium': color === 'medium',
         'telia-listItem--noBG': color !== 'dark' && color !== 'medium',
+        className,
       })}
     >
       <InnerTag
-        className={cn(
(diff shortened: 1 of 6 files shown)


src/business/List/ListItem.tsx
```diff
@@ -24,35 +24,28 @@ export const ListItem: React.FC<ListItemProps & ListStyle> = (props) => {
...
-        )}
+      <InnerTag
+        className={cn('telia-listItem__mainWrapper', {
+          'telia-listItem--compact': compact,
+          'telia-listItem--clickable': onClick,
+          className,
+        })}
         onClick={onClick}
-        onKeyDown={handleKeyDown}
         tabIndex={onClick && 0}
```
HeleneGrini, line 49: Dont need the tab index when we're using a button

HeleneGrini (changes requested):
Looks good! One small comment

HeleneGrini (approved):

PR #907, opened 2021-02-24
Make own useClickOutsideListener for sliding shopping cart

agnethesoraa:


src/organisms/SlidingShoppingCart/useClickOutsideListener.ts
```diff
@@ -0,0 +1,24 @@
+import { RefObject, useEffect } from 'react';
+
+export const useClickOutsideListener = (params: {
```
HeleneGrini, line 3: Perhaps put til in the `utils` folder and share it?

HeleneGrini: nvm i see you dont need the focus event!

agnethesoraa: Yes, the focus event seems to be creating some bugs in the cart. However, I figured that the other components might need that event, so I created a new click outside one for the sliding shopping cart without the focus event.

PR #908, opened 2021-02-24
Make possible additional paddding for the DropDownItem icon, if used

estien:
This one will allow for a more spacious layout when using icons in a dropdown. Will use it as this graphics shows
image

### DIFF
--- a/src/molecules/Dropdown/Dropdown.pcss
+++ b/src/molecules/Dropdown/Dropdown.pcss
@@ -111,7 +111,7 @@
   display: flex;
   align-items: center;
   color: var(--medium-dark-grey);
-  padding: 0.75rem 0.5rem;
+  padding: 1rem 0.95rem;
   margin: 0;
   font-size: 14px;
   background-color: white;
@@ -170,9 +170,12 @@
       font: inherit;
     }
   }
-  &__additional-icon-padding {
-    padding-left: 0.375rem;
+  &__icon-container {
     padding-right: 0.375rem;
+  }
+
+  &__sub-label {
+    padding-top:0.3rem;
   }
 }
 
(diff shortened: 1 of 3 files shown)


src/molecules/Dropdown/Dropdown.pcss
```diff
@@ -170,6 +170,13 @@
       font: inherit;
     }
   }
+  &__icon-container {
+    padding-right: 0.375rem;
+  }
+
+  &__sub-label {
```
HeleneGrini, line 177: We can use children without label in which case we dont want this padding. I think this should be the responsibility of the child, not something the dropdown controls 🤔

src/molecules/Dropdown/DropdownItem.tsx
```diff
@@ -26,6 +26,11 @@ export type DropdownItemProps = {
    */
   icon?: IconDefinition;
 
+  /**
+   * Give some additional horizontal padding around the icon if it is used
+   */
+  extraIconPadding?: boolean;
```
HeleneGrini, line 32: dont need this anymore ✂️

src/molecules/Dropdown/DropdownItem.tsx
```diff
@@ -101,13 +106,13 @@ export const DropdownItem: React.FC<DropdownItemProps> = (props) => {
...
       {props.icon ? (
-        <div>
+        <div className="telia-dropdown-item__icon-container">
           <Icon icon={props.icon} />
         </div>
       ) : null}
       <div>
         {props.label ? <div>{props.label}</div> : null}
-        <div>{props.children}</div>
+        {props.children ? <div className="telia-dropdown-item__sub-label">{props.children}</div> : null}
```
HeleneGrini, line 115: ref my other comment this is not necessarily a sub label -- it could be the entire content

HeleneGrini (changes requested):

HeleneGrini (approved):

PR #909, opened 2021-02-24
Enlarge icon in dropdown, in conjunction with the increased padding

estien:
From 
image
to 
image

### DIFF
--- a/src/molecules/Dropdown/Dropdown.pcss
+++ b/src/molecules/Dropdown/Dropdown.pcss
@@ -120,8 +120,8 @@
   text-align: left;
   width: 100%;
   .Icon {
-    width: 1rem;
-    height: 1rem;
+    width: 1.35rem;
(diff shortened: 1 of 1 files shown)


src/molecules/Dropdown/Dropdown.pcss
```diff
@@ -120,8 +120,8 @@
   text-align: left;
   width: 100%;
   .Icon {
-    width: 1rem;
-    height: 1rem;
+    width: 1.35rem;
```
HeleneGrini, line 123: From the sketches in the Design System it's supposed to be 1rem. If we want it bigger the sketches should be updated  

Screenshot 2021-02-24 at 12 14 27

estien: You have "steel control" @HeleneGrini  👍  I'll check with @BjornEvenEgge

PR #917, opened 2021-02-26
Add phone number component

mariannemagnussen:


src/molecules/PhoneNumberField/PhoneNumberField.stories.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+const countryCodes: CountryCode[] = [
+  { label: '+47 Norge', value: '+47' },
+  { label: '+46 Sverige', value: '+46' },
+  { label: '+358 Finland', value: '+358' },
+];
+
+export const Default = () => {
+  const [number, setNumber] = React.useState('');
+  console.log(number);
```
HeleneGrini, line 20: ✂️

src/molecules/PhoneNumberField/PhoneNumberField.tsx
```diff
@@ -0,0 +1,62 @@
...
+  maxlength?: number;
+  onChange: (number: string) => void;
+  placeholder?: string;
+  disabled?: boolean;
+  label?: string;
+} & ({ error: true; helpText?: string } | { error?: boolean; helpText?: string });
+
+export const PhoneNumberField = (props: PhoneNumberFieldProps) => {
+  const [landCode, setLandCode] = React.useState(props.countryCodes[0].value);
+  const [number, setNumber] = React.useState('');
```
HeleneGrini, line 25: i think we should support controlling this state from outside, and get all changes, not just the full number on onblur. Perhaps just have native change behaviour unless theres is sent handlers on props?

HeleneGrini (changes requested):
Looks good! It seems that there should be a purple border when the menu is open? 
Would also be nice with some stories of disabled and error 😁

HeleneGrini (approved):
nice 👍

PR #919, opened 2021-03-01
Fix/misc

mariannemagnussen:


src/molecules/QuickSearch/QuickSearch.pcss
```diff
@@ -17,7 +17,7 @@
 
   .telia-textfield-content {
     border-radius: 100px;
-    border: 1px solid var(--grey-200);
+    border: 1px solid var(--grey-200) !important;
```
HeleneGrini, line 20: hm. Do we need important or can we be more specific with the selectors?

mariannemagnussen: Yes, thats actually possible, thanks!

src/molecules/Dropdown/Dropdown.pcss
```diff
@@ -105,6 +105,17 @@
...
   }
+
+  &__disabled {
+    pointer-events: none;
+    background-color: var(--grey-100);
+    color: var(--grey-400);
+    border-color: var(--grey-300);
+  }
+
+  &__outline.telia-dropdown-toggle__purple {
```
HeleneGrini, line 116: should this also have `telia-dropdown-toggle__disabled` ?

mariannemagnussen: good catch

HeleneGrini (approved):

PR #920, opened 2021-03-01
Create FieldInstructions component

mariannemagnussen:


src/molecules/FieldInstructions/FieldInstructions.pcss
```diff
@@ -0,0 +1,132 @@
+.telia-field-instructions {
+  position: relative;
+
+  &--icon {
+    cursor: pointer;
+  }
+
+  &--card {
```
HeleneGrini, line 8: Seems that the font size should be 14 in the whole card, and the card header is too big?

src/molecules/FieldInstructions/FieldInstructions.stories.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { TextField } from '../TextField';
+import { set } from 'lodash/fp';
+
+export default {
+  component: FieldInstructions,
+  title: 'Component library/Molecules/Field Instructions',
+  decorators: [withDesign],
+};
+
+export const Default = () => {
```
HeleneGrini, line 13: Perhaps also add a story of a standalone FieldInstructions component ?  And I think this story should also be added to the TextField story. That's where I would look if I wanted a textfield with additional help/instructions (nice to have it here as well)

src/molecules/FieldInstructions/FieldInstructions.tsx
```diff
@@ -0,0 +1,50 @@
...
+  position?: 'top-start' | 'top' | 'top-end' | 'left' | 'right' | 'bottom-start' | 'bottom' | 'bottom-end';
+  /**
+   * Handle open/close state. Current usage is closing instructions onBlur for TextField.
+   * Want open/close on hovering? See Tooltip component
+   */
+  open: boolean;
+  setOpen: (open: boolean) => void;
+};
+
+export const FieldInstructions = (props: FieldInstructionsProps) => {
```
HeleneGrini, line 29: Should this be able to control the open state itself if its not provided from props? I guess you wouldn't always need to know the state or handle it explicitly

src/molecules/FieldInstructions/FieldInstructions.pcss
HeleneGrini: Seems that the button is too big, and I think that has happened a few times before. Perhaps we need another variant to our buttons/avatar?  One with less space.  Or if we just set the height of the header to match the sketches and let the button take up all the space? 🤷

HeleneGrini (changes requested):
Looks good! A few comments

src/molecules/FieldInstructions/FieldInstructions.stories.tsx
mariannemagnussen: Good point!

src/molecules/FieldInstructions/FieldInstructions.pcss
mariannemagnussen: Hmm..   Another variant is probably best since its been a problem before. Buuut im guessing we still want the ball-hovering effect.. so maybe solve it by letting the button set the paddings? 😅

src/molecules/FieldInstructions/FieldInstructions.tsx
mariannemagnussen: Maybe. When I thought about it I concluded that for those cases you might as well use the tooltip. But maybe not

src/molecules/FieldInstructions/FieldInstructions.pcss
HeleneGrini: Since the button already takes inn classname, we could/should perhaps just use that? For this component we could do it in the css?

mariannemagnussen: I dont really see how we can do it without remove the hovering effect?

HeleneGrini: What do you mean? 

Eg. setting 
``` 
    padding: 0;
    width: 20px;
    height: 20px;
```
on the button doesn't remove the hover effect, just makes the area smaller?

mariannemagnussen: True, but do you think this should be a variant of the button? so we pass a prop in and set something like that as style?

mariannemagnussen: Or should we just override the css in the stylesheet for this component?   I guess it looks better to minimize the hovering circle for the button

HeleneGrini: We should as UX about this -- if we should create a new variant for the button

mariannemagnussen: Asked Ingvild about this. She said that I could just use the button ball..  Maybe just leave it like this, and make updates later if they want changes?

HeleneGrini: Okei, cool. Fine by me! 😄

HeleneGrini (approved):

PR #922, opened 2021-03-02
change additional content size

hannaharnett:
Make additional content slightly larger and change the breaking point to xxl.

Change the menu breaking point to lg and add ellipsis if the menu item overflows.


antons1 (approved):

HeleneGrini (approved):
LGTM

PR #926, opened 2021-03-03
Use dedicated css classname instead of mixing with Icon component styling. Reduce cascading

estien:
This change makes it possible and/or cleaner to use the dropdown item with e.g. a leading icon without its styling to collide with what the styleguide dictates.


HeleneGrini (approved):
nice 👌

PR #938, opened 2021-03-11
refactor step indicator

HeleneGrini:
Started refactoring to extend with support for other labels inside of a step. Turns out we dont need it after all. Anyways, I finished converting it to tsx  🚶‍♀️


mariannemagnussen (approved):
lgtm!

PR #946, opened 2021-03-22
fix period picker css

HeleneGrini:


HeleneGrini (commented):
update with BEM pattern for classnames

mariannemagnussen (approved):
lgtm 👌

PR #954, opened 2021-03-26
Make type of title and description in StepFlow ReactNode to support customising content

HeleneGrini:


mariannemagnussen (approved):
Nice 👌

PR #957, opened 2021-04-08
Feature/design tweaks business

mariannemagnussen:
TextField:
- Update paddings
- Help text is left by default
- Height for default is 44px, height for compact is 36 px

Chip:
- Update paddings
- Size is set to 32px for all
- Square kind added
- previous kind (white/grey) is now called color. and kind is used to set rounded or square (BREAKING CHANGE!!)
- Added active colors

Checkbox:
- Add larger hovering and clickable area for checkbox - similar to radio btns
- Add active colors

PhoneNumberField:
- Height to 44px
- Added compact size
- fixed some border colors (due to updates in Textfield)
- set fixed width to fit 3 digits in country code 

RadioButton:
- Add active style

### DIFF
--- a/VERSIONS.md
+++ b/VERSIONS.md
@@ -15,20 +15,28 @@ For Major and Minor changes in version you must notify everyone in the #stylegui
 
 Only noteworthy changes.
 
+## 1.41.0
+
+- Updated Chip component
+  - New size (32px for all), size prop is removed
+  - Square chip is added
+  - Kind prop is renamed to color
+  - Change classnames to fit BEM pattern
+- Updated TextField
+  - Help text is left by default
+- Updated PhoneNumberField
+  - Change classnames to fit BEM battern
+  - Added size prop
+
 ## 1.40.0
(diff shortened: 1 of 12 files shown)


HeleneGrini (commented):
Very nice 👌 

I decided to test in other browser and the chip is stretching in FireFox. (not a new issue, but might as well fix it here now that you're fixing stuff 😄 ) 

Be sure to update the minor version with a remark on the breaking changes. Chips kind has changed, size is removed and classnames have changed (to follow BEM 👌 )

VERSIONS.md
```diff
@@ -15,20 +15,28 @@ For Major and Minor changes in version you must notify everyone in the #stylegui
 
 Only noteworthy changes.
 
+## 1.41.0
```
HeleneGrini, line 18: Burde ikke legge til dette før vi har satt ut egentlig, men la gå haha 😁

HeleneGrini (approved):

PR #981, opened 2021-05-06
Fix bug when crumbs array is undefined

remibengtbouvet:
Fix bug when crumbs prop is undefined and null. Made crumb[] nullable.

### DIFF
--- a/VERSIONS.md
+++ b/VERSIONS.md
@@ -17,6 +17,7 @@ Only noteworthy changes.
 ## 1.41.10
 - Added Breadcrumbs component
 - Added inputRef to TextField component
+- Change crumbs[] to nullable
 
 ## 1.41.0
 
--- a/VERSIONS.md
+++ b/VERSIONS.md
@@ -17,7 +17,6 @@ Only noteworthy changes.
 ## 1.41.10
 - Added Breadcrumbs component
 - Added inputRef to TextField component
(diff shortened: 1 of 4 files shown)


VERSIONS.md
```diff
@@ -17,6 +17,7 @@ Only noteworthy changes.
 ## 1.41.10
 - Added Breadcrumbs component
 - Added inputRef to TextField component
+- Change crumbs[] to nullable
```
oyvindbergbouvet, line 20: Remove this one - no need, none using breadcrumbs yet

src/molecules/BreadCrumbs/BreadCrumbs.stories.tsx
```diff
@@ -20,6 +24,8 @@ export const TenBreadCrumbsHideRoot = () => <BreadCrumbs crumbs={breadCrumbsData
 export const TenBreadCrumbsPageSizeLargeAndHideRoot = () => (
   <BreadCrumbs crumbs={breadCrumbsData} pageSize={4} alwaysShowRootCrumb={false} />
 );
+export const EmptyBreadCrumbsArray = () => <BreadCrumbs crumbs={[]} />;
+export const BreadCrumbsUndefined = () => <BreadCrumbs />;
```
oyvindbergbouvet, line 28: Looks good

PR #984, opened 2021-05-07
Feature/double data subscription

agnethesoraa:


src/molecules/SubscriptionLinesAccordion/SubscriptionLinesAccordion.jsx
```diff
@@ -189,7 +239,7 @@ const SubscriptionLinesAccordion = ({
             </div>
           </div>
         </div>
-        <div className="subscription-lines-accordion__description2 subscription-lines-accordion__mobile-only">
+        <div className="subscription-lines-accordion__description2 subscription-lines-accordion__mobile-only double-data">
```
joakikr, line 242: Should we prefix with `subscription-lines-accordion__double-data` ?

src/molecules/Subscription/Subscription.d.ts
```diff
@@ -47,6 +47,7 @@ declare interface SubscriptionProps {
   id?: string;
   name?: string;
   dataAmount?: SubscriptionDataAmount;
+  doubleDataAmount?: SubscriptionDataAmount;
```
joakikr, line 50: This is not needed from what i see 👍

joakikr (approved):

PR #985, opened 2021-05-07
Bugfix/add defaultprops breadcrumbs

remibengtbouvet:
Add deffault props to handle passing model from .net backend. If empty array is passed - default props should be used.


oyvindbergbouvet (approved):

remibengtbouvet:
Fix bug when backend pass model to frontend and component fails witth type = undefined

PR #994, opened 2021-05-14
Feature/breadcrumbs bkg color

remibengtbouvet:
Breadcrumb can set custom background color and not only inherit from parent. Font and icons can also get custom color.

### DIFF
--- a/src/molecules/Breadcrumbs/Breadcrumbs.jsx
+++ b/src/molecules/Breadcrumbs/Breadcrumbs.jsx
@@ -106,9 +106,11 @@ const Breadcrumbs = (props) => {
     return (
       <>
         <button type="button" className="telia-breadcrumbs__paging-button" onClick={onPagingEvent}>
-          <MoreIcon className="telia-breadcrumbs__more-icon" />
+          <MoreIcon className="telia-breadcrumbs__more-icon" style={SetMoreIconColor(props.colors)} />
         </button>
-        {showArrowLeftIcon && <ArrowLeftIcon className="telia-breadcrumbs__arrow-left-icon" />}
+        {showArrowLeftIcon && (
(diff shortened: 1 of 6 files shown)


src/molecules/Breadcrumbs/Breadcrumbs.stories.tsx
```diff
@@ -26,6 +26,36 @@ export const TenBreadcrumbsPageSizeLargeAndHideRoot = () => (
...
+  <Breadcrumbs colors={backgroundPurple} crumbs={breadCrumbsData.filter((item, i) => i <= 5)} />
+);
+export const GrayBackgroundColor = () => (
+  <Breadcrumbs colors={backgroundGray} crumbs={breadCrumbsData.filter((item, i) => i <= 5)} />
+);
+export const BlackBackgroundColor = () => (
+  <Breadcrumbs colors={backgroundBlack} crumbs={breadCrumbsData.filter((item, i) => i <= 5)} />
+);
+
+const backgroundPurple = {
```
oyvindbergbouvet, line 39: Change to color varaibles:
$core-purple instead of rgb hex, so it is defined in one place...
Styleguide already have those vars defined somewhere, so gues syou can simply write $core-purple without including anyhting

src/molecules/Breadcrumbs/Breadcrumbs.jsx
```diff
@@ -138,13 +140,29 @@ const Breadcrumbs = (props) => {
     );
   };
 
+  const SetBackgroundColor = (colors) => {
+    if (colors != null && colors.backgroundColor != null) return { backgroundColor: colors.backgroundColor };
```
oyvindbergbouvet, line 144: Hm this is correct, but isnt there a simpler, shorter way, to what we want? :/

PR #996, opened 2021-05-20
Feature/breadcrumbs bkg color

remibengtbouvet:
Update setting background styling  - styling now uses classNames().

### DIFF
--- a/src/molecules/Breadcrumbs/Breadcrumbs.jsx
+++ b/src/molecules/Breadcrumbs/Breadcrumbs.jsx
@@ -1,8 +1,6 @@
-import classNames from 'classnames';
 import React, { useState } from 'react';
 import { ArrowLeftIcon } from '../../atoms/Icon/icons';
 import { MoreIcon } from '../../atoms/Icon/icons';
-import classnames from 'classnames';
 
 const crumbType = {
   LINK: 'link',
@@ -24,6 +22,11 @@ const Breadcrumbs = (props) => {
   const maxCrumbIndex = props.crumbs.length - 1;
   const [maxIndex, setMaxIndex] = useState(maxCrumbIndex);
   const showLeftPagingTreshold = alwaysShowRootCrumb ? 1 : 0;
+
+  const colors = {
+    backgroundColor: props?.backgroundColor ?? null,
+    detailColor: props?.detailColor ?? null,
+  };
 
   let pagingSize = props.pagingSize ?? 1;
   let pageSize = props.pageSize ?? 3;
@@ -49,6 +52,13 @@ const Breadcrumbs = (props) => {
       type: crumbType.LINK,
     };
   });
+
+  const getStyle = (style, color) => {
+    if (color) {
+      return (style += ' ' + style + '--' + color);
+    }
+    return style;
+  };
 
   const getVisibleCrumbs = () => {
     return crumbs.filter((c) => {
@@ -98,7 +108,7 @@ const Breadcrumbs = (props) => {
 
   const CrumbLabel = ({ crumb }) => {
     return (
-      <span className={'telia-breadcrumbs__crumb-label'}>
+      <span className={getStyle('telia-breadcrumbs__crumb-label', colors.detailColor)}>
         <b>{crumb.name}</b>
       </span>
     );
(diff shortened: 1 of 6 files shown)


src/molecules/Breadcrumbs/Breadcrumbs.stories.tsx
```diff
@@ -26,6 +26,12 @@ export const TenBreadcrumbsPageSizeLargeAndHideRoot = () => (
 );
 export const EmptyBreadcrumbsArray = () => <Breadcrumbs crumbs={[]} />;
 export const BreadcrumbsUndefined = () => <Breadcrumbs />;
+export const PurpleBackgroundColor = () => (
+  <Breadcrumbs backgroundColor={'core-purple'} crumbs={breadCrumbsData.filter((item, i) => i <= 5)} />
```
oyvindbergbouvet, line 30: Core-purple? as text?
The value should come as a prop.BackgroundColor....

src/molecules/Breadcrumbs/Breadcrumbs.pcss
```diff
@@ -1,20 +1,40 @@
 .telia-breadcrumbs {
+    &--purple {
+        background-color: var(--core-purple);
```
oyvindbergbouvet, line 3: Doesnt styleguide have "Background-color" class "on its own", as in a global modifier?

So it was sufficient to add "background-color--core-purple" as class, instead of nesting it as a "child" of telia-breadcrumbs--purple?

PS: purple must be changed to core-puprle

src/molecules/Breadcrumbs/Breadcrumbs.jsx
```diff
@@ -23,6 +23,11 @@ const Breadcrumbs = (props) => {
   const [maxIndex, setMaxIndex] = useState(maxCrumbIndex);
   const showLeftPagingTreshold = alwaysShowRootCrumb ? 1 : 0;
 
+  const colors = {
```
oyvindbergbouvet, line 26: Remove this, just call the getStyle with the props.backColor 
- props are never null within a component

src/molecules/Breadcrumbs/Breadcrumbs.pcss
```diff
@@ -1,27 +1,99 @@
...
+    &--gray {
+        background-color: var(--grey-400);
+    }
+
+    &--white {
+        background-color: var(--white);
+    }
+
+    &--black {
+        background-color: var(--black);
```
oyvindbergbouvet, line 15: Can we have black background?
- When I think of it, we actually have black backgorund on some accordion header, good thinking!

src/molecules/Breadcrumbs/Breadcrumbs.pcss
```diff
@@ -1,27 +1,99 @@
...
+        &--black {
+            color: var(--black);
+        }
+
+        &--gray {
+            color: var(--gray-400);
+        }
+
+        @media all and (max-width: 40rem) {
+            padding-inline-start: 10px;
```
oyvindbergbouvet, line 34: padding-inline-start?
padding-left? or do you mean padding-right?
- i know that it gets "morphed" into either right or left, depending on how its going to be rendered...

src/molecules/Breadcrumbs/Breadcrumbs.pcss
```diff
@@ -1,27 +1,99 @@
...
+        &--black {
+            color: var(--black);
+        }
+
+        &--gray {
+            color: var(--gray-400);
+        }
+
+        @media all and (max-width: 40rem) {
+            &:not(:nth-last-child(-n + 2)) {
```
oyvindbergbouvet, line 55: pageSize and pagingSize probably adjusts the "-n+2"?
- what if pageSize is 10 (amount of items shown per Page)
- and pagingSize is 5 (if you click "next", how many Pages are moved forward)

src/molecules/Breadcrumbs/Breadcrumbs.stories.tsx
```diff
@@ -26,6 +26,20 @@ export const TenBreadcrumbsPageSizeLargeAndHideRoot = () => (
...
 export const BreadcrumbsUndefined = () => <Breadcrumbs />;
+export const PurpleBackgroundColor = () => (
+  <Breadcrumbs
+    backgroundColor={'core-purple'}
+    detailColor={'white'}
+    crumbs={breadCrumbsData.filter((item, i) => i <= 5)}
+  />
+);
+export const GrayBackgroundColor = () => (
+  <Breadcrumbs backgroundColor={'gray'} detailColor={'white'} crumbs={breadCrumbsData.filter((item, i) => i <= 5)} />
```
oyvindbergbouvet, line 37: Why hardcoded "gray" "white", why isnt that from props?

src/molecules/Breadcrumbs/Breadcrumbs.jsx
```diff
@@ -1,6 +1,8 @@
+import classNames from 'classnames';
 import React, { useState } from 'react';
 import { ArrowLeftIcon } from '../../atoms/Icon/icons';
 import { MoreIcon } from '../../atoms/Icon/icons';
+import classnames from 'classnames';
```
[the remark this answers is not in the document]
remibengtbouvet: Done

PR #1002, opened 2021-05-25
animation on hover

klounErik:


src/molecules/HardwareProductBox/HardwareProductBox.pcss
```diff
@@ -11,6 +11,7 @@
   color: var(--black);
   text-align: center;
   cursor: pointer;
+  transition: all 0.3s;
```
joakikr, line 14: This makes the text unfocused till transition is done. Consider adding only to `box-shadow` or remove the scaling on hover?

PR #1018, opened 2021-06-03
removed whitespace under accordion

klounErik:


joakikr (changes requested):
Not looking good on list of accordions:

Screenshot 2021-06-03 at 14 08 07

joakikr (approved):

PR #1048, opened 2021-06-28
Fix/date picker

HeleneGrini:
- improving setting of selected dates
- add placeholder props
- enable handling open state


mariannemagnussen (approved):
👍

PR #1085, opened 2021-10-01
Feature/footer add api example for getting footer data from telia no

remibengtbouvet:


package.json
```diff
@@ -121,6 +121,7 @@
         "@fortawesome/free-brands-svg-icons": "^5.15.4",
         "@fortawesome/react-fontawesome": "^0.1.15",
         "archiver": "^5.0.0",
+        "axios": "^0.21.4",
```
oyvindbergbouvet, line 124: Do we need axios - instead of just fetch?

PR #1114, opened 2021-11-10
Add opacity layer for shopping cart accordion

agnethesoraa:


src/organisms/ShoppingCartAccordion/ShoppingCartAccordion.pcss
```diff
@@ -1,9 +1,20 @@
...
+}
+
+.shopping-cart__opacity-layer {
+  width: 100vw;
+  height: 100vh;
+  position: absolute;
+  top: 0;
+  left: 0;
+  z-index: 100;
+  background: rgba(0, 0, 0, 0.6);
```
joakikr, line 17: The opacity layer does not show in Gatsby (due to parent being sticky and not relative positioning)
Screenshot 2021-11-19 at 09 59 47

Tried changing `position: fixed`, which shows the opacity layer but does not look good on with the menu:
Screenshot 2021-11-19 at 10 00 03

Changing to the following looks better:

- .shopping-cart__sticky: z.index back to 10
- .shopping-cart__opacitu-layer: z-index set to 9
- Adding background-color white for menu

Screenshot 2021-11-19 at 10 03 53

joakikr (commented):
Also, think you need to change node back to 12 until we have updated it on our apps. The package lock changes causes trouble, so they need to be reverted (they are for version 2) 😄

src/organisms/ShoppingCartAccordion/ShoppingCartAccordion.pcss
agnethesoraa: 👍

PR #1116, opened 2021-11-17
Feature/card add option for skipping purple underline

remibengtbouvet:


oyvindbergbouvet (approved):
Look OK (I didt not testrun though)

PR #1131, opened 2022-01-05
Accessibility update

hannaharnett:
* Blocks background scrolling
* Blocks tabbing outside modal on first tab
* Return focus to original trigger or a specified html element
* add role of dialog
* add aria-labelledby as a required prop in Modal


mariannemagnussen (approved):
Looks good 👌

PR #1136, opened 2022-01-17
Add totalMva & totalPriceMVA rows to ShoppingCart

joakikr:


asmundth (approved):

asmundth:
Should we use VAT in stead of MVA in the names?

PR #1150, opened 2022-02-07
Fix font-family to menu dropdown row

agnethesoraa:
Without this the dropdown row looks slightly "lighter" in boldness compared to the other menu elements as can be seen in the screensot below.


agnethesoraa:
Screenshot 2022-02-07 at 10 25 56

joakikr (approved):

PR #1160, opened 2022-03-11
Override apple-system-blue for default html-tag value - set to black

remibengtbouvet:
For iOS button has color -apple-system-blue. When running styleguide components on iOS default color is not black. Explicitly set color to black.

### DIFF
--- a/src/atoms/Button/Button.pcss
+++ b/src/atoms/Button/Button.pcss
@@ -13,7 +13,6 @@
 }
 
 .button {
-  color: black;
   background-color: var(--white);
   border: 1px solid var(--core-purple);
   border-radius: 2rem;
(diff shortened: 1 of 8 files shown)


src/atoms/Button/Button.pcss
```diff
@@ -13,6 +13,7 @@
 }
 
 .button {
+  color: black;
```
joakikr, line 16: Is this necessary? Already have `color: var(--core-purple);` in this class styling

joakikr (approved):
LGTM - just one small comment

src/atoms/Button/Button.pcss
remibengtbouvet: If you open https://telia.no/priser-i-utlandet you will se that the dropdown menu has blue font. It seems that button for ios has apple-system-blue as default color. It might be that var(--black) ia better but for now it will be blue on iOS if it is not forced.

PR #1164, opened 2022-03-16
remove z- index on toggle thumb

hannaharnett:
Removed the z-index because it shows up on top of fixed footer when scrolling. Doesn't seem to be necessary to function as it should, but want to check if it was added for a specific reason? 😅


mariannemagnussen (approved):
looks good to me 😄

PR #1207, opened 2022-05-12
New font for headings

agnethesoraa:


src/global.pcss
```diff
@@ -1,87 +1,96 @@
...
+*,
+:after,
+:before {
+  box-sizing: inherit;
 }
 
 @font-face {
-    font-family: 'pebble';
-    src: url('{{assetPath}}/fonts/Pebble_W_Rg.woff') format('woff');
+  font-family: 'pebble';
```
oyvindbergbouvet, line 19: Whats this? Why is there a 'pebble' remaining? :)

agnethesoraa: Pebble is not removed from everything yet. It was decided to start with headers because as Kurt said it was the most important to get done before Monday. The other components will be changed in later PRs.

klounErik (approved):

PR #1213, opened 2022-05-31
add optional ref to TableBodyRow 

hannaharnett:
Added optional refs so that we can manage focus between our list/table and details card view


CyrielleAlbert (approved):
lgtm!

PR #1223, opened 2022-06-22
Explicit children

niklasholm:
The implicit children property on components are deprecated and are being removed in React 18. Some type libs might adapt to this change without a major version bump, possibly breaking dependents (@types/styled-components is an example of this).

Add explicit children properties to those components that need it now to avoid this and prepare for future support of React 18.

See also:
- https://solverfox.dev/writing/no-implicit-children
- DefinitelyTyped/DefinitelyTyped#56210


package.json
```diff
@@ -67,7 +67,7 @@
         "@types/lodash": "^4.14.144",
         "@types/pretty": "^2.0.0",
         "@types/react": "^17.0.44",
-        "@types/react-dom": "^18.0.3",
+        "@types/react-dom": "^17.0.16",
```
oyvindbergbouvet, line 70: What? Was it 18? Oh well...

oyvindbergbouvet (approved):

PR #1253, opened 2022-09-29
change z-index of SpeechBubbleTabs so arrow wouldn't hide if parent c…

adrianasabanaite:
…ontainer has background

### DIFF
--- a/src/molecules/SpeechBubbleTabs/SpeechBubbleTabs.pcss
+++ b/src/molecules/SpeechBubbleTabs/SpeechBubbleTabs.pcss
@@ -39,6 +39,8 @@
         }
 
         &--selected {
+            cursor: default;
+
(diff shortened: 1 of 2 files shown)


oyvindbergbouvet (approved):

src/molecules/SpeechBubbleTabs/SpeechBubbleTabs.pcss
```diff
@@ -52,7 +54,10 @@
                 background-color: #30004B;
                 border-bottom: 1px solid #30004B;
                 border-right: 1px solid #30004B;
-                z-index: -1;
+            }
+
+            span {
```
oyvindbergbouvet, line 59: Approved - but you dont like to put & before span? :)

PR #1263, opened 2022-10-17
New prop - autoSetStepFromUrl

oyvindbergbouvet:
basic logic to find the initial step based on current path only (ignores query strings for now)

### DIFF
--- a/src/molecules/StepIndicatorPaging/Data/StepDataChildrenAsHtmlWithUrls.tsx
+++ b/src/molecules/StepIndicatorPaging/Data/StepDataChildrenAsHtmlWithUrls.tsx
@@ -40,7 +40,7 @@ export const StepDataChildrenAsHtmlWithUrls = () => {
     },
     {
       title: 'Step Url 6 Long title text here',
-      url: '#Step6',
+      url: '/iframe.HtmL?q=1000',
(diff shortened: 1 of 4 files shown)


src/molecules/StepIndicatorPaging/Data/StepDataChildrenAsHtmlWithUrls.tsx
```diff
@@ -40,7 +40,7 @@ export const StepDataChildrenAsHtmlWithUrls = () => {
     },
     {
       title: 'Step Url 6 Long title text here',
-      url: '#Step6',
+      url: '/iframe.HtmL?q=1000',
```
oyvindbergbouvet, line 43: Checking that "case insensitive" works ("HtmL")

src/molecules/StepIndicatorPaging/Models/Props.tsx
```diff
@@ -1,13 +1,13 @@
 import { Step } from './Step';
 
 export interface Props {
-  steps?: Step[] | null | undefined;
+  steps?: Step[] | null;
```
oyvindbergbouvet, line 4: Question marks is already saying "undefined" is allowed

PR #1287, opened 2022-12-05
Use BreadcrumbsProps already defined in file.

k1rd3rf:
We're getting unnecessary warning when using this component, and this component already handles optional props.

image

### DIFF
--- a/src/molecules/Breadcrumbs/Breadcrumbs.tsx
+++ b/src/molecules/Breadcrumbs/Breadcrumbs.tsx
@@ -1,6 +1,6 @@
 import React, { useState } from 'react';
-import { ChevronRightIcon } from '../../atoms/Icon/icons';
-import { MoreLowIcon } from '../../atoms/Icon/icons';
+import { ChevronRightIcon } from '../../atoms/Icon';
+import { MoreLowIcon } from '../../atoms/Icon';
 
 export interface Crumb {
   name: string;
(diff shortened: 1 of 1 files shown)


src/molecules/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -37,15 +37,7 @@ const arrowRight = {
   NONE: 'none',
 };
 
-const Breadcrumbs = (props: {
-  crumbs: Crumb[];
```
k1rd3rf, line 41: defined on line 18

src/molecules/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -91,14 +83,9 @@ const Breadcrumbs = (props: {
...
-  const getStyle = (style: string, color: string) => {
-    if (!color) {
-      color = 'black';
-    }
-    return style + ' ' + style + '--' + color;
-  };
+  const getStyle = (style: string, color?: string) => style + ' ' + style + '--' + (color ? color : 'black');
 
-  const setBackgroundColor = (style: string, color: string) => {
+  const setBackgroundColor = (style: string, color?: string) => {
```
k1rd3rf, line 88: Made color optional here, since BreadcrumbsProps had set as optional

joakikr (approved):

PR #1293, opened 2022-12-13
feat(table): add empty state

tomas-anikevicius:


mariannemagnussen (approved):
Nice 👌  
You could actually add the props like this:
```
&
  (
    | {
        tableIsEmpty: boolean;
        emptyTableLabel: string;
      }
    | {}
  );
```
Then you would not be able to send in only one of them. But you don't have to do that.

PR #1316, opened 2023-03-13
Upgrade to React 18

espegj:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@telia/styleguide",
-  "version": "1.48.32",
+  "version": "1.49.0",
```
k1rd3rf, line 3:
```suggestion
  "version": "1.50.0",
```

espegj: I see that there is some automatic to set this version, so I removed this change. But I still use `1.49.0` in VERSIONS.md why `1.50.0`?

leskraas (approved):

PR #1340, opened 2023-06-02
update color, hover styles and remove active styles on radio btn according to VOCA

hannaharnett:
- Changed color from core-purple to black-purple
- Changed hover styles to only show thicker border (no shadows) and only apply to enabled radio btn 
- Removed active effects


asmundth:
Too old => close it

PR #1354, opened 2023-07-27
carousel component

adrianasabanaite:


src/molecules/Carousel/Carousel.pcss
```diff
@@ -0,0 +1,104 @@
+.carousel {
```
joakikr, line 1: Rename with prefix to avoid namespace collides for consumers: `.telia-carousel` ?

src/molecules/Carousel/Carousel.pcss
```diff
@@ -0,0 +1,104 @@
...
+                    cursor: pointer;
+                    background-color: #6E6E6E;
+                    border-radius: 50%;
+                    opacity: 50%;
+                }
+            }
+        }
+    }
+
+    .pages-indicator {
```
joakikr, line 74: These does not center with slides if screen is bigger than slides
(At least not in the storybook provided)

src/molecules/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,157 @@
...
+              >
+                {currentCarouselPage.map((item, index) => (
+                  <HardwareProductBox key={item.name + item.brand + index} {...item} />
+                ))}
+              </div>
+            </div>
+          ))}
+        </div>
+        {totalPages > 1 && (
+          <div
```
joakikr, line 144: Should arrows also be accessible by keyboard navigation? Also it's not possible to swipe the cards on mobile.

joakikr (commented):
Design looks good 🎉 
Some comments related to naming and universal design.

joakikr (approved):

PR #1355, opened 2023-08-02
add border to accordion to make them pop

klounErik:


src/molecules/SubscriptionAccordion/SubscriptionAccordion.pcss
```diff
@@ -6,6 +6,7 @@
   border-radius: 3px;
   background-color: var(--white);
   box-shadow: 0px 1px 4px 0px rgba(0, 0, 0, 0.12);
+  border: 1px solid rgba(229, 229, 235, 1);
```
joakikr, line 9: Just curious, where did you get this rgba from? This translated to `#e5e5eb` but in designs I see `#AAAAB0`

joakikr (approved):

src/molecules/SubscriptionAccordion/SubscriptionAccordion.pcss
klounErik: After discussion with designers we agreed on the one that has #e5e5eb

PR #1358, opened 2023-08-22
carousel prices

adrianasabanaite:


src/molecules/Carousel/Carousel.tsx
```diff
@@ -158,7 +165,29 @@ export const Carousel: React.FC<CarouselProps> = ({ items }: CarouselProps) => {
...
                 {currentCarouselPage.map((item, index) => (
-                  <HardwareProductBox key={item.name + item.brand + index} {...item} />
+                  <HardwareProductBox key={item.name + item.brand + index} {...item}>
+                    <div>
+                      {item.priceDescription && (
+                        <div className="hardware-product-box__product-price-decription">{item.priceDescription}</div>
+                      )}
+                      {item.price && (
+                        <div className="hardware-product-box__product-price">
+                          {item.price},- <span>pr md.</span>
```
oyvindbergbouvet, line 175: Monthly price like this, probably not hardcode, suffix it...

THe app's using the carousel can instead hardcode the "priceSuffix" if it doesnt come from api whatever

PR #1359, opened 2023-09-21
business/add higher contrast and make hover more visible for black tabs

hannaharnett:
The black version should have a dotted underline on hover
The outline & black version should work as before


CyrielleAlbert (approved):
Lgtm 🥇

PR #1362, opened 2023-09-26
added footer for unexpanded accordion

klounErik:
image


alfredbirk (commented):
Nice! How about calling the prop just `footer` instead of `compactFooterChildren`? (Same name as in SubscriptionLinesAccordion)

alfredbirk (approved):

PR #1366, opened 2023-10-02
Feature/style adjustments for subscription compact accord

klounErik:
image
image

### DIFF
--- a/src/molecules/SubscriptionCompactAccordion/SubscriptionCompactAccordion.pcss
+++ b/src/molecules/SubscriptionCompactAccordion/SubscriptionCompactAccordion.pcss
@@ -188,8 +188,8 @@
   }
 
   &__icon-arrow {
-    width: 1rem !important;
-    height: 1rem !important;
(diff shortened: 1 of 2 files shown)


src/molecules/SubscriptionCompactAccordion/SubscriptionCompactAccordion.pcss
```diff
@@ -73,6 +74,10 @@
     padding: 1rem;
     margin: 0;
 
+    &--footer {
+      padding: 1.5rem 2rem 0rem !important;
+    }
+
     @media all and (min-width: 41em) {
       padding: 1.5rem 2rem;
     }
```
alfredbirk, line 83: If you put it after the other padding, then the important-tag is not needed.
```suggestion
    @media all and (min-width: 41em) {
      padding: 1.5rem 2rem;
    }

    &--footer {
      padding: 1.5rem 2rem 0rem;
    }
```

PR #1369, opened 2023-10-04
fix(Modal): disable clickOutsideListener in case of fullscreen Modal

CyrielleAlbert:
bugfix: On small screens, any click inside our outside the modal would close it, when it's fullscreen modal.
When size of the modal is fullscreen, we don't need that listener, so I disabled it.


hannaharnett (approved):
👏

PR #1374, opened 2023-10-13
add hover effect and change border radius

klounErik:
image


joakikr (commented):
Hover on regular looks good 🎉 
Hover on the purple variant need another hover color than dark I think

joakikr (approved):

PR #1378, opened 2023-10-18
Setting activeIndex to 'activeIndex' prop instead of selectedHeaderIn…

oyvindbergbouvet:
…dex which does nada, so it actually selects either "Private" or "Bedrift [Company]" at "page load"

Do not set focus if activeIndex has been set, else you end up with "focus" css rules, which are equal to active index rule, so both seems to be selected Added a story to verify the feature

### DIFF
--- a/src/molecules/Menu/Menu.jsx
+++ b/src/molecules/Menu/Menu.jsx
@@ -100,7 +100,9 @@ const Menu = ({
     const focusableElements = mobileMenuRef.current.querySelectorAll(focusableElementsSelector);
     if (!focusableElements.length) return;
 
-    focusableElements[0].focus();
+    if (activeIndex === 0) {
+      focusableElements[0].focus();
+    }
   };
 
   useEffect(setFocusOnFirstFocusableElement);
@@ -128,7 +130,7 @@ const Menu = ({
         LinkTemplate={LinkTemplate}
         onMobileMenuToggle={toggleMobileMenu}
         menuLinks={menuLinks}
-        selectedHeaderIndex={activeIndex}
+        activeIndex={activeIndex}
         onMenuItemSelected={toggleMobileMenu}
         isLoading={isLoading}
         mobileMenuCloseButtonLabel={mobileMenuCloseButtonLabel}
--- a/src/molecules/Menu/Menu.jsx
+++ b/src/molecules/Menu/Menu.jsx
@@ -100,7 +100,7 @@ const Menu = ({
     const focusableElements = mobileMenuRef.current.querySelectorAll(focusableElementsSelector);
     if (!focusableElements.length) return;
 
-    if (activeIndex === 0) {
+    if (!activeIndex || activeIndex === 0) {
       focusableElements[0].focus();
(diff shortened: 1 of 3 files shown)


src/molecules/Menu/Menu.jsx
```diff
@@ -100,7 +100,9 @@ const Menu = ({
     const focusableElements = mobileMenuRef.current.querySelectorAll(focusableElementsSelector);
     if (!focusableElements.length) return;
 
-    focusableElements[0].focus();
+    if (activeIndex === 0) {
+      focusableElements[0].focus();
+    }
```
oyvindbergbouvet, line 105: Do not set focus if activeIndex was set, then we end up with two "active index" (by the looks at least, as focus and active gets same css)

src/molecules/Menu/Menu.jsx
```diff
@@ -128,7 +130,7 @@ const Menu = ({
         LinkTemplate={LinkTemplate}
         onMobileMenuToggle={toggleMobileMenu}
         menuLinks={menuLinks}
-        selectedHeaderIndex={activeIndex}
+        activeIndex={activeIndex}
```
oyvindbergbouvet, line 133: This seems to been a bug, selectedHeaderIndex... 
- activeIndex already existed, so I just set it

src/molecules/Menu/Menu.stories.tsx
```diff
@@ -254,6 +254,118 @@ export const LoggedInDropdownMenuB2C = () => {
...
+    },
+  ];
+
+  return (
+    <Menu
+      loginUrl="#Menu"
+      logoUrl="#"
+      logoTitle="Telia logo"
+      menuLinks={menuLinks}
+      activeIndex={1}
```
oyvindbergbouvet, line 358: ActiveIndex 1 - equals the "second menu", BEdrift [Company] in our case
- as index 0 is "private"

src/molecules/Menu/MobileMenu/MobileMenu.jsx
```diff
@@ -52,10 +52,11 @@ const MobileMenu = ({
   menuLinks,
   onMenuItemSelected,
   isLoading,
+  activeIndex,
```
oyvindbergbouvet, line 55: Passing the activeIndex to mobile menu too

src/molecules/Menu/MobileMenu/MobileMenu.jsx
```diff
@@ -52,10 +52,11 @@ const MobileMenu = ({
...
   mobileMenuCloseButtonLabel,
   buttonValues,
 }) => {
-  const [selectedHeaderIndex, setSelectedHeaderIndex] = useState(0);
+  const [selectedHeaderIndex, setSelectedHeaderIndex] = useState(activeIndex ?? 0);
```
oyvindbergbouvet, line 59: Setting the initialState to activeIndex if specified, or 0

SimkeAsta (approved):

PR #1382, opened 2023-10-23
Feature/rounded borders subscription compact accordion

klounErik:


joakikr (commented):
Seems like cards are jumping slowly on expand/collapse compared to earlier:

https://github.com/TeliaSoneraNorge/styleguide/assets/5881281/968cc16d-9a9e-4de3-bca7-b16d3a6abb1d

PR #1436, opened 2023-12-04
StepIndicator: Don't refresh when changing step

alfredbirk:
https://trello.com/c/KDUSTnuz/3286-reswitch-starting-on-pdp-coming-back-from-minside-should-default-step-to-subscription


alfredbirk:
Was not needed.

PR #1450, opened 2024-01-02
make whole cart clickable

klounErik:


src/organisms/SlidingShoppingCart/useClickOutsideListener.ts
```diff
@@ -4,14 +4,15 @@ export const useClickOutsideListener = (params: {
...
-  const { open, close, ref } = params;
+  const { open, close, ref, buttonRef } = params;
 
   useEffect(() => {
     const handleClose = (e: MouseEvent) => {
+      const mobileOutside = e.target && buttonRef?.current && !buttonRef.current.contains((e.target as any) as Node);
       const targetIsOutside = e.target && ref?.current && !ref.current.contains((e.target as any) as Node);
-
-      if (open && targetIsOutside) {
+      if ((open && targetIsOutside) || (open && mobileOutside)) {
```
k1rd3rf, line 15:
```suggestion
      if (open && (targetIsOutside || mobileOutside)) {
```

k1rd3rf (approved):

alfredbirk (commented):
Would be nice if it was possible to collapse it by click anywhere on top when it is expanded (like an Accordion)

PR #1460, opened 2024-01-16
show upfront and monthly

klounErik:
image


src/organisms/SlidingShoppingCartV2/SlidingShopppingCartV2.pcss
```diff
@@ -75,7 +75,39 @@
...
+    display: flex;
+    flex-direction: column;
+    :nth-child(1) {
+      margin-bottom: .5rem;
+    }
+
+    .heading {
+       font-size: 1rem !important;
+       color: var(--core-purple-850) !important;
+    }
```
alfredbirk, line 109: I think you can avoid using `!important` if you instead add a new className to <Header>, like this:
```
<Heading className=...
```

PR #1498, opened 2024-04-02
expandable textcontainer

klounErik:
image
image


src/molecules/ExpandableTextContainer/ExpandableTextContainer.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+  return (
+    <Container
+      className={cn('telia-expandableTextContainer', {
+        ['telia-expandableTextContainer--expanded']: expanded,
+      })}
+    >
+      {textBlocks.map((block) => (
+        <div className="telia-expandableTextContainer__textblock">
+          <Heading tag="h1" size="xs">
```
alfredbirk, line 28: Should not be h1 since a page should only have one h1-tag. H2 is probably fine.

src/molecules/ExpandableTextContainer/ExpandableTextContainer.pcss
```diff
@@ -0,0 +1,42 @@
+.telia-expandableTextContainer {
+    height: 200px;
```
alfredbirk, line 2: In design sketches only the first 2.5 lines are shows. So maybe a bit less height when not expanded?

alfredbirk (approved):

PR #1656, opened 2025-05-22
fix: Remove light purple hover color

madistaimre:
CR fix for my initial commit where I added the styling in next project. Made it so that the purple and black ones do not lose their color when expanded and hovering (confirmed with Kristian)


alfredbirk:
Tested and it works great 👍  Just wondering, why did you add those `!important` properties? It seems to work without them as well

alfredbirk (approved):

